PR #123, opened 2026-08-25
Add App Store screenshots for iPhone and iPad

umeranjum17:
## What changed

- rebuilt split view around one navigator/content contract with Spaces, sessions, Tools, New session, Settings, and live content
- extracted shared compact/comfortable Spaces and live-herd models
- added responsive sidebar sizing, selection, stable replacement navigation, and aligned Back behavior
- made Live Terminals adaptive across split layouts and phones
- restored complete CLI-served local Web UI and reactive control/view-only authority
- applied measured responsive layout to declarative plugin screens
- captured and composed six truthful iPhone and six truthful iPad App Store screenshots from the simulator app

## App Store story

Both sets lead with New Agent, Files, and Terminal.

- iPhone continues with Runbook, Settings, and Appearance.
- iPad now continues with the actual Herd landing surface, Settings, and Realtime voice. This replaces the sparse Runbook/Appearance frames and shows the primary product surface.

The iPad captures use a deterministic Product Launch fixture with neutral `/Users/Shared/muxr-demo` paths, named Claude/Codex review tasks, populated terminal previews, the actual 1,532-file repository explorer, and no usernames, `/private/tmp`, model codenames, automation output, failures, or QA commentary.

Composition adds only the marketing background, headline, crop, border, and shadow. iPad logo/headline spacing now matches the iPhone system while preserving headline/device clearance.

## Verification

- iPad Pro 13-inch: New Agent, Files, terminal, Herd, Settings, and Realtime voice visually reviewed from final composites
- iPhone 17 Pro Max: existing six reviewed assets retained byte-for-byte
- `yarn workspace @muxr/mobile typecheck`
- `yarn vitest run apps/mobile/sources/utils/responsive.spec.ts apps/mobile/sources/utils/liveTerminalOrder.spec.ts` — **13 passed**
- `asc screenshots validate --path screenshots/app-store/en-US/iphone-6.9 --device-type IPHONE_69` — **6 ready, 0 errors, 0 warnings**
- `asc screenshots validate --path screenshots/app-store/en-US/ipad-13 --device-type IPAD_PRO_3GEN_129` — **6 ready, 0 errors, 0 warnings**

Latest screenshot remediation: `18a8d39`.

PR #120, opened 2026-08-24
Fix macOS and iOS simulator validation gaps

umeranjum17:
## Problem

The full local suite and native iOS simulator flow exposed platform-specific failures that prevented reliable macOS validation and left several simulator paths unusable:

- macOS `/var` versus `/private/var` path aliases broke plugin authority assertions and release bundling.
- Darwin uses a different PTY tool and symlink-removal behavior than Linux.
- Ad-hoc simulator builds have no Keychain entitlement, so startup failed closed before the app could be tested.
- The native preview bridge crashes in `react-native-tcp-socket` on recent iOS simulators.
- The new-agent screen routed before a newly started session reached the synchronized store.
- Stale Expo Router dev-screen declarations emitted warnings on every startup.

## Changes

- Canonicalize temporary/plugin paths and the Xcode bundle phase's `PROJECT_ROOT`.
- Drive Darwin TTY smoke flows through `/usr/bin/expect`; keep util-linux `script` on Linux.
- Stop fixture relays through their owned PID file when `/proc` is unavailable.
- Use `unlinkSync` for directory symlinks in extension lifecycle checks.
- Add an iOS-simulator-only AsyncStorage fallback for secrets; physical devices and Android remain on SecureStore.
- Bypass the TCP preview bridge on the iOS simulator and use the Mac's shared loopback directly.
- Poll session synchronization before the new-agent screen opens the terminal.
- Remove registrations for development routes that no longer exist.
- Add preview failure diagnostics and build-policy coverage.

## Simulator verification

Tested on an iPhone 16e simulator running iOS 26.3.1 with Xcode 26.3:

- Debug and arm64 Release native builds
- Cold startup, local account restoration, relay/host reconnect
- Settings, connection, voice-provider list, plugins, appearance themes, languages, preferences, changelog, machine details, pairing validation, notification permission
- App-launched OpenCode agent, live terminal render, terminal input, and confirmed agent stop/cleanup
- Real Herdr-backed terminal input (`SIMULATOR_OK` observed in the pane)
- Files/repository browser and README viewer
- Changed-files list and diff viewer
- Git history and Usage screens
- Runbook list
- Photo picker, attachment upload, prompt path propagation, and SHA-256 verification
- Project preview discovery and WebView load through a local HTTP fixture
- Realtime voice missing-provider handling
- Dictation failure handling when simulator audio input is unavailable

[...]

PR #124, opened 2026-08-25
refactor: apply thermonuclear codebase audit

umeranjum17:
## Problem

This is an end-to-end thermonuclear maintainability audit of the current repository, not a review of one feature diff. The audit covered the application runtimes (`mobile`, `host`, `relay`, `probe`), shared packages (`contract`, `crypto`, `wire`), bundled plugins, native mobile modules, operational/package scripts, and demo tooling: 593 source files and roughly 94k lines.

Priority order was structural correctness and ownership, then spaghetti/branching debt, then lower-impact cleanup. Risky security-sensitive decompositions remain explicitly listed below instead of being mixed into this PR without dedicated coverage.

### Structural correctness

- Fixed process-tree shutdown escalation: a previously-signalled child was incorrectly skipped during the later `SIGKILL` pass because `ChildProcess.killed` means “a signal was sent”, not “the process exited”.
- Made package lifecycle/smoke cleanup deterministic through relay pid files, canonical temporary paths, platform-stable process discovery, Linux-only TTY checks, symlink-safe cleanup, and bounded relay shutdown waits.
- Made the tracked/package secret scanner skip non-file symlink targets instead of trying to read a linked directory.
- Canonicalized plugin executable/provenance roots before authority comparison. This fixes `/var` versus `/private/var` aliases on macOS without weakening the pinned-root invariant.

### Canonical ownership

- Moved untrusted plugin-result bounding/sanitization into `@muxr/contract`; host execution and the plugin CLI now use one implementation.
- Replaced the host’s bespoke basename implementation with `node:path`.
- Removed an identity manifest-action wrapper.
- Removed dead server-configuration setters/fallbacks and the unreachable remote-log transport they kept alive.
- Centralized generated-avatar props and hashing across all four renderers, removing four copies and `ComponentType<any>`.
- Centralized reactive device-authority policy and migrated every control-producing mobile surface to the same fail-closed hook.

### Decomposition and dependency direction

- Extracted HomeDock environment/machine/project/worktree/agent policy into `useHomeDockEnvironment`.
  - `HomeDock.tsx`: 1,150 → 937 lines.
  - Removed three unused Zustand subscriptions, dead `supportsWorktree` branches, duplicate agent-correction effects, and a one-value picker dispatcher.
- Extracted declarative chart rendering and its Skia/Victory/animation dependencies.
  - `DeclarativeScreen.tsx`: 946 → 716 lines.
- Moved primitive, markdown, and native-menu contracts into type-only modules. The audited TS/TSX/MJS relative-import graph now has zero strongly connected components; it had three cycles before this change.
- Included the installed thermonuclear audit skill and lock metadata.

## Validation

- `yarn run check`: **30/30 checks passed**.
  - workspace and mobile typechecks
  - wire build
  - 165 Vitest flows
  - crypto, host-domain, contract, relay, and layout checks
  - pairing/revocation/relay/voice/preview/Herdr/worktree E2E flows
  - package install/lifecycle smokes
  - core-purity and secret policies
- Focused `pluginCatalog.test.ts`: **19/19 passed**.
- Import-cycle scan: **0 cycles** after the type-boundary changes.
- `git diff --check`: passed.

## Remaining prioritized audit findings

These are real findings, but combining them with this PR would materially increase correctness/security risk.

### P0 — host session source is one enormous mutable closure

`apps/host/src/herdr/herdrSessionSource.ts` is 1,973 lines; `createHerdrSessionSource` spans lines 253–1973. It owns discovery projections, lifecycle epochs, status watches, attention, attachments, snapshots, agent start/stop, plugin refresh, plugin calls, streams, and the returned `SessionSource` implementation. Next change should extract projection/lifecycle and plugin execution controllers behind the existing `SessionSource` contract, with current integration flows retained.

### P0 — relay composition and self-host HTTP authority are fused

`apps/relay/src/relay.ts` is 1,097 lines; `startRelay` spans lines 115–1062. Roughly 480 lines of enrollment, tickets, pairing, grants, device revocation, push, and static-web HTTP policy sit inline with WebSocket routing and process lifetime. This is security-sensitive. Decompose by authority domain only after adding flow coverage for each extracted endpoint group; a giant context-object move would merely relocate the complexity.

### P1 — local setup is an operational monolith

`scripts/local-setup.mjs` is 2,895 lines with about 545 branch points. Installation, service repair, daemon definitions, hosted auth, self-host state, ingress, pairing, device rotation, account commands, uninstall, and doctor repairs share one module. Split around the existing command entry points and preserve state formats/atomic-write ordering.

### P1 — mobile sync admission has avoidable head-of-line blocking

`PluginExecutionGate.drain` in `apps/mobile/sources/sync/sync.ts` returns when the queue head has reached its per-plugin cap, even when an unrelated plugin behind it could use free global capacity. Fix this together with a single flow-level scheduling test and consolidation of the adjacent global-only gate; changing rate-limit scheduling without coverage does not belong in this structural PR.

### P1 — terminal and Android voice surfaces remain oversized

- `TerminalScreen.tsx`: 935 lines with one component spanning lines 86–935; terminal channel lifecycle, sibling navigation, links, attachments, gesture state, tool menus, status, and rendering are fused.
- `VoiceOverlayService.kt`: 604 lines, including a 350-line companion object combining static bridge API, notification state, and service coordination. The foreground-service-before-mic invariant makes casual decomposition unsafe.

### P2 — provider adapters duplicate a runtime

The xAI, OpenAI, and Gemini streaming plugins repeat tool declarations, prompt policy, key loading, Herdr execution, client-frame handling, audio chunking, refusal parsing, and lifecycle plumbing. Their per-plugin-root installation boundary prevents a naive sibling import. A shared runtime needs an explicit packaging/isolation design rather than copy-paste cleanup in this PR.

## Scope discipline

No feature behavior, protocol, persistence format, voice provider policy, responsive asset, or UI vocabulary was intentionally changed. The PR deletes duplicate/dead ownership, establishes focused module boundaries, and fixes issues directly demonstrated by the repository suite.

### DIFF
--- a/.agents/skills/thermo-nuclear-code-quality-review/SKILL.md
+++ b/.agents/skills/thermo-nuclear-code-quality-review/SKILL.md
@@ -1,0 +1,192 @@
+---
+name: thermo-nuclear-code-quality-review
+description: Run an extremely strict maintainability review for abstraction quality, giant files, and spaghetti-condition growth. Use for a thermo-nuclear code quality review, thermonuclear review, deep code quality audit, or especially harsh maintainability review.
+disable-model-invocation: true
+---
+
+# Thermo-Nuclear Code Quality Review
+
+Use this skill for an unusually strict review focused on implementation quality, maintainability, abstraction quality, and codebase health.
+
+Above all, this skill should push the reviewer to be **ambitious** about code structure. Do not merely identify local cleanup opportunities. Actively search for "code judo" moves: restructurings that preserve behavior while making the implementation dramatically simpler, smaller, more direct, and more elegant.
+
+## Core Prompt
+
+Start from this baseline:
+
+> Perform a deep code quality audit of the current branch's changes.
+> Rethink how to structure / implement the changes to meaningfully improve code quality without impacting behavior.
+> Work to improve abstractions, modularity, reduce Spaghetti code, improve succinctness and legibility.
+> Be ambitious, if there is a clear path to improving the implementation that involves restructuring some of the codebase, go for it.
+> Be extremely thorough and rigorous. Measure twice, cut once.
+
+## Non-Negotiable Additional Standards
+
+Apply the baseline prompt above, plus these explicit review rules:
+
+0. **Be ambitious about structural simplification.**
+   - Do not stop at "this could be a bit cleaner."
+   - Look for opportunities to reframe the change so that whole branches, helpers, modes, conditionals, or layers disappear entirely.
+   - Prefer the solution that makes the code feel inevitable in hindsight.
+   - Assume there is often a "code judo" move available: a re-organization that uses the existing architecture more effectively and makes the change dramatically simpler and more elegant.
+   - If you see a path to delete complexity rather than rearrange it, push hard for that path.
+
+1. **Do not let a PR push a file from under 1k lines to over 1k lines without a very strong reason.**
+   - Treat this as a strong code-quality smell by default.
+   - Prefer extracting helpers, subcomponents, modules, or local abstractions instead of letting a file sprawl past 1000 lines.
+   - If the diff crosses that threshold, explicitly ask whether the code should be decomposed first.
+   - Only waive this if there is a compelling structural reason and the resulting file is still clearly organized.
+
+2. **Do not allow random spaghetti growth in existing code.**
+   - Be highly suspicious of new ad-hoc conditionals, scattered special cases, or one-off branches inserted into unrelated flows.
+   - If a change adds "weird if statements in random places", treat that as a design problem, not a stylistic nit.
+   - Prefer pushing the logic into a dedicated abstraction, helper, state machine, policy object, or separate module instead of tangling an existing path.
+   - Call out changes that make the surrounding code harder to reason about, even if they technically work.
+
+3. **Bias toward cleaning the design, not just accepting working code.**
+   - If behavior can stay the same while the structure becomes meaningfully cleaner, push for the cleaner version.
+   - Do not rubber-stamp "it works" implementations that leave the codebase messier.
+   - Strongly prefer simplifications that remove moving pieces altogether over refactors that merely spread the same complexity around.
+
+4. **Prefer direct, boring, maintainable code over hacky or magical code.**
+   - Treat brittle, ad-hoc, or "magic" behavior as a code-quality problem.
+   - Be skeptical of generic mechanisms that hide simple data-shape assumptions.
+   - Flag thin abstractions, identity wrappers, or pass-through helpers that add indirection without buying clarity.
+
+5. **Push hard on type and boundary cleanliness when they affect maintainability.**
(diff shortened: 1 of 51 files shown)


umeranjum17:
Review verdict: **BLOCK — do not resolve this as one giant conflict and merge.** The individual cleanup work is mostly sound, but this branch predates merged PR #120, overlaps heavily with #123, and has contradictory authority policy.

### Required before any merge

1. **Split the PR.** Recommended lanes:
   - low-risk ownership/dedupe: `boundRpcDisplay`→contract, dead server/log deletion, `atomicWriteJson`, type-only modules, avatar dedupe, `up.mjs` kill escalation
   - responsive extractions after #123: HomeDock + DeclarativeScreenChart
   - plugin provenance canonicalization as its own security-reviewed change
   - drop the vendored audit skill/symlink/lockfile from the product repo

2. **Drop `useDeviceAuthority` and its consumer migrations from this PR.** PR #123 already owns and fixes this exact policy. This version hard-codes local web to `observe`, contradicting the local host’s full-control boundary and re-breaking New Agent/composer/sidebar controls. It also lacks `.catch()` on `loadHostedGrant()`, so a storage/JSON failure leaves every control permanently loading/disabled with an unhandled rejection.

3. **Rebase onto `eb018d1c` and preserve PR #120’s macOS `/usr/bin/expect` TTY coverage.** The current `checkPackageSmoke.mjs` side gates TTY tests to Linux and would silently remove the new macOS fullscreen/prompt/menu/update cancellation checks. Keep main’s Darwin harness; only layer genuinely new cross-platform improvements on top.

4. **Require the actual CI suite.** GitHub has only CodeQL/analysis checks; no `suite` job ever ran for this head. The local 30/30 claim was against the pre-#120 tree and is not a merge gate.

5. **Remove or justify the provenance relaxation.** Canonicalizing `expectedRoot` fixes the real `/var`↔`/private/var` issue. Canonicalizing the untrusted `parsed.root` before equality changes an exact stored-root pin into alias equivalence. Keep the former; split and explicitly test/justify the latter.

Also note the fixed provenance identity changes `manifestHash` on affected macOS installs, causing a safe but user-visible one-time plugin re-approval; that belongs in release notes rather than under “no behavior changed.”

Verified clean in isolation: child SIGKILL escalation, RPC bounding dedupe, HomeDock/chart lifts, dead remote-log removal, basename replacement, avatar hash centralization, and type-boundary moves.

umeranjum17:
Superseded per review; this monolithic branch must not be merged.

Replacement lanes rebuilt on current main (`eb018d1`):
- #125 — low-risk ownership/dedupe only; excludes authority, responsive, provenance, skill, and package-smoke changes
- #126 — plugin provenance canonical-root fix with exact stored-root pinning and explicit re-approval release note

`useDeviceAuthority` was dropped in favor of #123's policy. HomeDock/DeclarativeScreen responsive extractions are deferred until #123 merges so they can be rebuilt without conflicts. The vendored audit skill/symlink/lockfile were dropped.

PR #125, opened 2026-08-25
refactor: centralize shared ownership contracts

umeranjum17:
## Summary

Low-risk ownership and deduplication lane split from #124 after review. Rebuilt directly on `eb018d1` / current `main`; excludes responsive work, device-authority policy, provenance changes, audit skill files, and package-smoke changes from the superseded branch.

- move plugin result bounding/sanitization into `@muxr/contract` so host and CLI share one implementation
- delete dead server configuration setters and unreachable remote logging transport
- use `node:path.basename` instead of a bespoke host helper
- remove the manifest row-action identity wrapper
- centralize generated-avatar props/hash and remove `ComponentType<any>`
- move markdown, primitive, and native-menu contracts into type-only modules, eliminating the detected relative-import cycles
- allow process shutdown escalation to signal a still-running child even after an earlier signal was sent

## Scope

No HomeDock/chart extraction and no `useDeviceAuthority` migration. Those overlap #123 and are intentionally absent. No plugin provenance semantics changed.

## Validation

- `yarn run check` — **30/30 passed** on current main
- `yarn workspace @muxr/mobile typecheck` — passed
- `git diff --check` — passed

Supersedes the low-risk portion of #124.

### DIFF
--- a/apps/host/src/domain/atomicWriteJson.ts
+++ b/apps/host/src/domain/atomicWriteJson.ts
@@ -1,5 +1,5 @@
 import { mkdir, rename, writeFile } from 'node:fs/promises';
-import { dirname, join } from 'node:path';
+import { basename, dirname, join } from 'node:path';
 
 /** Atomic JSON write: temp file in the same directory, then rename. */
 export async function atomicWriteJson(filePath: string, value: unknown): Promise<void> {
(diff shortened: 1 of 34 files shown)


umeranjum17:
Follow-up review: **ready to merge.** This is the clean low-risk split requested from #124.

Verified behavior-preserving: shared RPC bounding is byte-equivalent and remains mobile-safe; dead server/log paths were truly unreachable; avatar hashes/defaults do not drift; type-only boundary moves retain every consumer; the manifest wrapper was identity-only; and `up.mjs` correctly allows SIGKILL escalation after SIGTERM.

CI/CodeQL are green, `diff --check` passes, and synthetic merges with #126 and #123 are clean in either order.

PR #126, opened 2026-08-25
fix(host): canonicalize plugin provenance root

umeranjum17:
## Summary

Security-reviewed provenance fix rebuilt on current `main`.

On macOS and other symlink-parent layouts, the configured extensions directory can use one path while `realpath` returns another. Host validation and `muxr package` now canonicalize the already-owner-checked extensions root before containment comparison, so host and package list/update/remove agree on npm ownership.

The stored provenance `root` remains an **exact string pin** against the canonical executable root; it is not canonicalized or relaxed.

## Security invariants

- extensions and `.provenance` must still be real directories, not symlinks
- plugin root must still be exactly one child named by plugin id
- provenance files still use `O_NOFOLLOW`, size bounds, and schema validation
- `parsed.root !== executableRoot` still fails closed
- a same-directory spelling such as `canonicalRoot/.` is explicitly tested and rejected
- package removal reuses the same canonical ownership check before deleting materialization

## Coverage

- platform-independent symlink-parent host provenance flow
- package install → list → update → remove through an aliased parent
- materialization and provenance removal assertions
- exact stored-root mismatch rejection

## Release note

Affected macOS installs whose prior approval hash was calculated through a `/var` alias may see a safe one-time plugin re-approval because the corrected source identity can change the manifest approval hash.

## Validation

- `yarn vitest run apps/host/src/herdr/pluginCatalog.test.ts` — **19/19 passed**
- `node scripts/packageLifecycleSmoke.mjs` — passed, including aliased lifecycle
- `yarn run check` — **30/30 passed** on exact rerun; the first run hit the unrelated attachment-watcher timing flake and its focused rerun passed
- `git diff --check` — passed

Supersedes the provenance portion of #124.

### DIFF
--- a/apps/host/src/herdr/pluginCatalog.test.ts
+++ b/apps/host/src/herdr/pluginCatalog.test.ts
@@ -214,17 +214,24 @@ describe('plugin catalog flow', () => {
         expect(() => catalog.action(changed.pluginId, changed.manifestHash!, 'missing')).toThrow('unavailable or changed');
 
         const muxrHome = await realpath(await mkdtemp(join(tmpdir(), 'muxr-home-')));
+        const aliasParent = await realpath(await mkdtemp(join(tmpdir(), 'muxr-home-alias-')));
+        const muxrHomeAlias = join(aliasParent, 'home');
+        await symlink(muxrHome, muxrHomeAlias, 'dir');
         const extensionRoot = join(muxrHome, 'extensions');
         const managedRoot = join(extensionRoot, 'example.muxr-ui');
         await mkdir(managedRoot, { recursive: true });
         await writeFile(join(managedRoot, 'muxr-ui.json'), JSON.stringify({ schemaVersion: 1, pluginId: 'example.muxr-ui', contributions: [] }));
         await mkdir(join(extensionRoot, '.provenance'));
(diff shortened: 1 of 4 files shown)


umeranjum17:
Security review: the host-side change is minimal and preserves the exact `parsed.root` pin, symlink guards, one-child containment, `O_NOFOLLOW`, and size/schema checks. **One root-cause blocker remains before merge.**

`scripts/package.mjs:81-89` `validRoot()` performs the same mismatched comparison:

```js
const real = realpathSync(root);
const expected = resolve(base);
relative(expected, real);
```

After this PR, the host recognizes an aliased `/var`→`/private/var` install as npm while `muxr package list/update/remove` still treats it as local. Consequences include failed npm updates and incomplete removal that leaves materialization/provenance behind.

Canonicalize the already-owner-checked `base` there too, then add a platform-independent symlink-parent flow covering host + package list/update/remove agreement. The current macOS-only string rewrite does not exercise the fix in Linux CI.

umeranjum17:
Fixed the remaining root-cause blocker in `c242eaf`.

- `scripts/package.mjs` now canonicalizes the already-owner-checked extensions base in `validRoot()`.
- `validNpmProvenance()`, `isNpmRoot()`, and npm removal use the same canonical expected child root.
- `parsed.root` remains an exact string comparison; it is not canonicalized.
- Added a platform-independent symlink-parent host flow plus package install/list/update/remove coverage. The test asserts both materialization and provenance are removed.

Focused host tests and lifecycle smoke pass. Full `yarn run check` passed 30/30 on the exact rerun.

PR #127, opened 2026-08-25
feat: add cross-machine agent collaboration

umeranjum17:
## Summary

- add Settings-mediated, capability-scoped collaboration between paired computers
- add signed peer authorization, directed E2EE, durable idempotency, expiry, rotation, revocation, repair, and offline convergence
- let allowed machines list/read/status/watch/prompt agents, with constrained peer tools for native realtime voice
- pin voice to its original machine/session while refreshing only that machine's grant on reconnect
- fix configured machine names and neutral inactive-pairing status

## Security and recovery

- five-minute mutation freshness and 64 receipts per peer
- security-first revocation bypasses attacker-controlled receipt admission
- crash-recoverable authorization journal and orphan cleanup
- fresh correlated liveness proof after every socket open
- strict decrypted-frame validation and null-safe error handling
- no raw Herdr CLI, close-pane, shell, takeover, or arbitrary plugin authority in peer voice tools
- hosted authority uses existing generic pair-session/device/grant/rotation APIs; self-host keeps dedicated peer routes

## Verification

- `node scripts/runSuite.mjs` — **30/30 passed** on final HEAD
- focused peer/mobile/voice flows — **16/16 passed**
- all three voice provider lifecycle/syntax checks — passed
- workspace and Expo/React Native typechecks — passed
- `git diff --check` — passed
- hostile `gpt-5.6-sol` security/correctness review — all 8 blockers repaired
- final Claude Fable 5 oracle review — **no release blockers**

## Residual deployment check

The hosted route sequence matches the contract-tested generic control-plane API. A live hosted two-machine ceremony remains the final post-deploy smoke check; no private repository was changed by this PR.


umeranjum17:
Thermonuclear review verdict: **BLOCK**.

The green suite does not cover several broken ownership paths. Independent high-rigor review and direct source verification found these blockers:

### P1 — completion watches never return to the requesting peer

`apps/host/src/peer/runtime.ts:681-688` returns immediately after registering `agent.watch`. Settlement later exists only as broadcast `watch.settled`, while `NodePeerClient` accepts correlated `result` frames only (`apps/host/src/peer/client.ts:206-240`) and peer grants do not carry the broadcast data key. The source voice/tool caller can only receive `{ watching: true }`, never completion. Add a directed peer settlement result/event path and have the broker consume it.

### P1 — the host-owned voice broker is not a capability boundary

`apps/host/src/peer/broker.ts:174-190` accepts unauthenticated JSON on the predictable owner-only socket. Every plugin runs as the same OS user and receives `HOME`/`MUXR_HOME`, so any plugin can derive and connect to it. The request is cast rather than parsed; an unknown method with `text` falls through to the prompt branch (`broker.ts:108-141`). Pass an unguessable per-stream capability or connected private FD to the approved voice child, and strictly validate the request union before dispatch.

### P1 — one transient startup failure permanently disables collaboration

`apps/host/src/main.ts:538-561` couples durable recovery and optional broker startup in one try/catch, then discards `peerRuntime` on any failure. A temporary relay/control-plane outage during pending revocation means the journal cannot recover when connectivity returns; all administration stays unavailable until daemon restart. Attach the runtime first, fence mutations while recovery is pending, retry recovery, and isolate broker startup failure from collaboration administration.

### P1 — mobile intent can forget live host authority

`apps/mobile/sources/collaboration/computerCollaboration.ts:236-258` only filters/updates edges already present in AsyncStorage. It never imports relationships returned by canonical host `peer.list`. If local intent is cleared/corrupt or a second controller opens Settings, UI reports Off while peer credentials remain live; reconnect creates duplicates and later Disconnect revokes only newly recorded edges. Rebuild intent from matching host inbound/outbound relationships and surface unmatched halves as repair-needed.

### P2 — ambiguity has no addressable resolution

`apps/host/src/peer/broker.ts:94-105,155-159` detects duplicate machine/agent names but exposes no stable human selector. Every clarification attempt repeats the same ambiguity error unless an internal id leaks. Persist unique user-facing machine aliases and return qualified agent aliases that subsequent calls can resolve exactly.

### P2 — split the 861-line peer state machine before extending it

`apps/host/src/peer/runtime.ts:83-861` owns authorization journaling, install, key-rotation revocation, outbound routing/aliases, and mutation receipts in one class. Keep `PeerRuntime` as a typed dispatcher; extract a crypto transition coordinator, receipt executor, and outbound peer service so each durable recovery state machine can be audited independently.

Deployment gap remains: the PR still lacks the live hosted two-machine ceremony needed to verify grant refresh/rotation, reconnect, and immediate socket revocation.

umeranjum17:
Owner review resolved in commits `9a66f944`, `5c645fef`, and `7b47f31a`.

- watch settlement is private/correlated/directed and durable across reconnect
- prompt/watch retries retain stable operation IDs; causal plugin ACK prevents duplicate delivery
- broker uses a closed request union and per-stream capability; revoke aborts admitted sockets/controllers
- runtime recovery retries without detaching administration; revocation fences local ingress first
- mobile reconstructs and normalizes canonical `peer.list` authority, including unmatched halves
- aliases persist and remain human-readable/non-secret through churn
- outbound and receipt responsibilities were extracted from `PeerRuntime`

Enabled plugin backends remain explicitly trusted, unsandboxed same-user code; the stream capability is least-ambient routing, not a claim of hostile-local-process isolation.

Verification: final suite **30/30**, exact-head CI/CodeQL green. Final Claude Fable 5 whole-stack review: **no release blockers; PR #127 READY**.

Remaining post-deploy check: live hosted two-machine grant rotation/reconnect/immediate-revocation ceremony.

PR #151, opened 2026-08-28
feat: make the herd glanceable and addressable

umeranjum17:
## Summary
- keeps one full-size horizontal terminal strip in Agent Route creation order across working, blocked, done, failed, and stopped states
- keeps tree-backed cards visible while the session catalog join lags; removes card caps, expiry, counters, Session history, paths, and provider/context duplication
- makes Task Title primary and canonical Agent Name secondary across Herd presentation, accessibility copy, activity records, and tests
- freezes settled terminal preview polling while retaining the final frame
- adds persisted “While you were away” acknowledgements for unseen blocked/done/failed transitions from the last 24 hours, latest per Agent Route, collapsed to three rows with bounded overflow
- acknowledges activity on tap or only after one second with the Herd screen focused, the app foregrounded, and the complete terminal card visible horizontally and vertically
- omits lifecycle ages when the session timestamp has not joined yet

## Validation
- `yarn vitest run apps/mobile/sources/herd/application/liveTerminalOrder.spec.ts apps/mobile/sources/herd/domain/herd.spec.ts apps/mobile/sources/catalog/application/sessionSync.integration.spec.ts --config apps/mobile/vitest.config.ts` — 10 passed
- `yarn workspace @muxr/mobile typecheck` — passed
- `yarn vitest run apps/mobile/sources/architecture.spec.ts --config apps/mobile/vitest.config.ts` — 6 passed
- `git diff --check` — passed
- terminology gate under `apps/mobile/sources/herd/**`: no `Human Name` or `humanName` matches


apps/mobile/sources/herd/presentation/LiveTerminalsRow.tsx
```diff
@@ -1,257 +1,230 @@
...
+            const end = start + CARD_WIDTH;
+            return start >= scrollX && end <= scrollX + stripWidth ? [card.id] : [];
+        }));
+        const eventIds = activityRows
+            .filter((row) => visibleSessionIds.has(row.sessionId))
+            .map((row) => row.eventId);
+        if (eventIds.length === 0) return;
+        const timer = setTimeout(() => markSeen(eventIds), 1000);
+        return () => clearTimeout(timer);
+    }, [activityRows, cards, markSeen, scrollX, stripWidth]);
```
umeranjum17: **Blocking — auto-ack fires while the user cannot see the strip.** This effect gates only on horizontal scroll math inside the strip. It has no screen-focus gate (`useFocusEffect`/`useIsFocused` — the established pattern one directory over in `useHerdTreeLive.ts:40`), no vertical-visibility check, and no AppState check. Two concrete failure modes:

1. HerdView stays mounted beneath pushed screens. Tap a completion notification into a session screen, and every other agent's blocked/done/failed event whose card sits in the strip's default viewport is marked seen ~1s after it arrives — While You Were Away is empty when the user returns. That inverts the feature's contract ("unseen … acknowledged after one second of *full card visibility*").
2. The strip is the list header; scroll the session list down and the horizontally-"visible" cards keep acking while vertically offscreen.

`TerminalPreview` already gates its polling on AppState; this timer needs at least a focus gate before the acks persist.

apps/mobile/sources/herd/domain/Agent.ts
```diff
@@ -1,12 +1,12 @@
...
 
 /**
- * The person-shaped row in the Herd. Agent Route authorizes; Human Name and
+ * The terminal-shaped card in the Herd. Agent Route authorizes; Agent Name and
  * Task Title are display-only and never used as keys.
  */
 export type AgentRecord = {
     route: string;
-    humanName: string;
+    agentName: string;
```
umeranjum17, line 9: **Blocking — invents a synonym the glossary bans.** `CONTRIBUTING.md:85-89`: "Names in code match CONTEXT.md: … Human Name, Task Title … Do not invent synonyms." `CONTEXT.md:41-43` defines **Human Name** and explicitly lists "agent name" under _Avoid_. This rename (`humanName` → `agentName`, "Agent Name" in doc comments here and in `agentPresentation.ts`) lands without touching `CONTEXT.md`, `CONTEXT-MAP.md:17`, `apps/mobile/sources/catalog/CONTEXT.md:15`, or `packages/README.md:35` — so the repo now carries two names for one concept, and the new one is the banned synonym. Meanwhile `catalog/` (`humanNameFromHost`, `humanNameForNotice`) and `watch/` keep the old term for the same value (`metadata.displayName`). If Agent Name is the intended canonical term, rename it in the glossary and the aligned docs/code in this same change; otherwise keep Human Name.

apps/host/src/agent/infrastructure/layoutSelfCheck.ts
```diff
@@ -196,6 +197,8 @@ async function demo(): Promise<void> {
     const first = await ask(socketPath, access.capability, { method: 'prompt', agent: 'John', text: 'Keep going.', operationId: 'op-0' });
     const replayed = await ask(socketPath, access.capability, { method: 'prompt', agent: 'John', text: 'Keep going.', operationId: 'op-0' });
     assert(first.ok === true && replayed.ok === true && prompts === 1, 'an accepted operation id replays without rerunning the mutation');
+    const taskStatus = await ask(socketPath, access.capability, { method: 'status', agent: 'Harden audio' });
+    assert(taskStatus.data === 'John is idle.', 'a unique Task Title resolves to its Human Name');
```
umeranjum17: This newly added assertion message says "resolves to its Human Name" while the mobile half of this same diff renames that concept to Agent Name. Whichever term wins the glossary decision (see `Agent.ts` comment), the diff should not introduce both vocabularies for the same concept.

plugins/voice/coordinatorPolicy.mjs
```diff
@@ -4,15 +4,31 @@ import { createConnection } from 'node:net';
 const agentProperty = {
     agent: {
         type: 'string',
-        description: 'Spoken human agent name. Omit only for the active agent.',
+        description: 'Human Name, Task Title, or provider kind. Omit only for the active agent.',
```
umeranjum17: Same vocabulary split: this diff adds new "Human Name" strings here (tool description, `list_agents` description, coordination instructions) while renaming the concept to "Agent Name" in mobile. Pick one term and align with CONTEXT.md in this change.

apps/mobile/sources/herd/domain/herd.ts
```diff
@@ -102,12 +103,17 @@ export function sortHerd(sessions: Session[], workspaces: readonly HerdrTreeWork
                 name: pane.displayName?.trim() || nameOf(session, paneDisplayName(pane)),
                 taskTitle: pane.taskTitle?.trim() || session?.metadata?.taskTitle?.trim() || paneTaskTitle(pane),
                 status: pane.agentStatus,
+                changedAt: session?.metadata?.lifecycleStateSince ?? session?.updatedAt ?? 0,
```
umeranjum17: Non-blocking: for a tree pane whose session hasn't joined yet, `changedAt` falls back to `0`, and `agentStateLabel(status, 0)` renders `Needs you · 20000d`-style epoch ages on blocked/done/failed/idle cards — in exactly the catalog-lag scenario this PR is built to survive. Suggest treating `0` as unknown and omitting the age suffix (as `agentAccessibilityLabel` already does for `changedAt === undefined`).

apps/mobile/sources/herd/domain/herdTree.ts
```diff
@@ -15,34 +15,18 @@ export function hasAgent(ws: HerdrTreeWorkspace): boolean {
...
-        tabs: workspace.tabs.map((tab) => ({
-            ...tab,
-            agentStatus: 'unknown',
-            panes: tab.panes.map((pane) => ({ ...pane, agentStatus: 'unknown' })),
-        })),
-    }));
+/** A transport blip does not rewrite the last observed agent lifecycle. */
+export function lifecycleTree(workspaces: readonly HerdrTreeWorkspace[], _connected?: boolean): HerdrTreeWorkspace[] {
+    return workspaces as HerdrTreeWorkspace[];
 }
```
umeranjum17: Non-blocking: `lifecycleTree` is now the identity function with a dead `_connected` parameter, yet `useHerdTreeLive.ts:12` and `KernelNotifications.tsx:37` still pass `socketStatus === 'connected'` and keep `socketStatus` in their memo deps — recomputing/re-sorting on every status flip for nothing. Clean cutover: drop the parameter, update the three callers (or delete the function and use the workspaces directly).

apps/mobile/sources/herd/presentation/LiveTerminalsRow.tsx
```diff
@@ -1,257 +1,230 @@
...
-        () => sortHerd(sessions, lifecycleTree(workspaces, socketStatus === 'connected')),
-        [sessions, socketStatus, workspaces],
+    const [scrollX, setScrollX] = React.useState(0);
+    const handleLayout = React.useCallback((event: LayoutChangeEvent) => setStripWidth(event.nativeEvent.layout.width), []);
+    const panes = React.useMemo(() => herdPanes(sessions, lifecycleTree(workspaces)), [sessions, workspaces]);
+    const cards = React.useMemo(() => {
+        const next = reconcileLiveTerminalCards(cardsRef.current, selectLiveTerminalCards(sessions, panes));
+        cardsRef.current = next;
+        return next;
+    }, [panes, sessions]);
```
umeranjum17: Non-blocking: mutating `cardsRef.current` inside `useMemo` is a render-phase side effect — a discarded concurrent render still leaks its ordering into the ref. It happens to be benign here because `reconcileLiveTerminalCards` is idempotent for a given `current`, but the conventional shape (reconcile in an effect, or `useState` + updater) would not rely on that.

umeranjum17 (commented):
**Verdict: REQUEST CHANGES** (submitted as a comment review — GitHub refuses the request-changes event from the PR author's own account).

Reviewed the full diff at 550dd6d (against merge-base a861490) against the stated requirements and repo standards (CONTRIBUTING.md, CONTEXT.md, apps/mobile AGENTS.md). Reran the touched flow tests locally at the PR head: `liveTerminalOrder.spec.ts` + `herd.spec.ts` + `sessionSync.integration.spec.ts` (9 passed), `providerRefusal.spec.mjs` (8 passed), `architecture.spec.ts` (6 passed).

**Requirements that check out**
- Fixed 208×152 cards in every lifecycle state; MAX_CARDS/grid/expiry removed; `reconcileLiveTerminalCards` keeps surviving slots and appends new panes in creation order — no shrinking/reordering (`liveTerminalOrder.ts:47-69`).
- Tree panes stay visible through a lagging catalog join: `selectLiveTerminalCards` no longer drops panes without a session (`liveTerminalOrder.ts:20-37`), covered by the new flow test.
- Task Title primary / Agent Name secondary across `LiveTerminalCard`, `SpacesTree`, `RecentActivity`, and accessibility labels via `agentPresentation.ts`.
- Session history buttons removed from `HerdView`/`SidebarView`; path/provider/branch duplication removed from card footers; `getSessionSubtitle` reduced to the canonical name; `useAttentionRows` no longer leaks session-id prefixes.
- Settled polling freezes with the final frame retained (`TerminalPreview` does one `read()` when `live === false`, no interval); other `TerminalPreview` callers keep prior behavior via the optional props.
- While You Were Away: blocked/done/failed only, 24h window, latest per Agent Route (seen-latest suppresses older unseen — tested), 3 collapsed rows + bounded `+N more` (domain cap 8), persisted acks (AsyncStorage, bounded to 128 ids, serialized writes), tap = ack + scroll-or-navigate, hides at zero rows. `lifecycleEvents` is newest-first (`boundLifecycleEvents` sorts desc), so the latest-per-route scan is sound; host `catalog()` likewise returns newest-first for the voice `activity` tool.
- `apps/mobile/sources/herd/**` is clean of `humanName`/`Human Name`/`animalName`.

**Blocking findings** (inline)
1. The one-second acknowledgement effect has no screen-focus/visibility gate, so unseen blocked/done/failed activity is marked seen — and persisted — while the user cannot see the strip. This defeats the feature's core promise.
2. Vocabulary split: the diff introduces "Agent Name" in mobile while CONTEXT.md still canonizes "Human Name" and explicitly lists "agent name" under _Avoid_ — and the same diff adds fresh "Human Name" prose in host/voice files. CONTRIBUTING.md: "Names in code match CONTEXT.md … Do not invent synonyms." The glossary/docs must move in the same change, one term everywhere.

Three non-blocking notes inline (epoch-age labels for lagging joins, `lifecycleTree` dead parameter, ref mutation inside `useMemo`). Also note the branch forked before #152 (`bb937a44`); it merges cleanly since this PR doesn't touch `agentKinds()`, but a rebase before merge would be prudent.

Requesting changes on findings 1–2.

umeranjum17 (commented):
**Verdict: APPROVE** — final integrated re-review at exact HEAD `4482bed7b18ecfc6309319471ddf2217872dfe08` (merge-base with main `a8614903`). Submitted as a comment review because GitHub refuses the approve event from the PR author's own account; treat this as the explicit approval. **No blocking findings.**

## Prior blockers — both resolved
1. **Acknowledgement visibility (was blocking).** The 1s ack is now gated on `useIsFocused()` + AppState foreground + a fire-time `measureInWindow` re-check, feeding the pure `visibleActivityEventIds` (`liveTerminalOrder.ts:98-113`) which requires focused ∧ foreground ∧ strip fully inside the vertical viewport (`visibilityTop`/`visibilityBottomInset` passed by all three callers: HerdView:273-277/329-332, MainView:521) ∧ card fully inside the horizontal viewport (`LiveTerminalsRow.tsx:131,139-141,159-163,180-207`). Verified by unit test ("acknowledges only fully visible cards on a focused foreground Herd screen") and by the API 36 emulator gate: background no-ack, offscreen no-ack, partial-visibility no-ack (>2s), full-visibility ack at ~1.6s, ack persisted across app restart.
2. **Vocabulary (was blocking).** #153/#156 finish the cutover: CONTEXT.md now canonizes **Agent Name** (the user-editable Herdr `agent.name`), CONTRIBUTING/CONTEXT-MAP/catalog+watch CONTEXT/READMEs updated in the same branch; host Human-Name pool deleted (identity schema v4→v5), `normalizeAgentName`/`parseAgentName` in contract with a `pp_`/`pph_` internal-id guard; wire field stays `displayName` with a documented compatibility note. Repo-wide grep for `humanName|Human Name|animalName|humanKey`: **zero matches**.

My three prior non-blocking notes are also fixed: epoch-age labels (`changedAt` now optional; `agentStateLabel` omits age when unknown, tested at `liveTerminalOrder.spec.ts:68`), `lifecycleTree` deleted with all three callers cleaned up, and `reconcileLiveTerminalCards` now returns the previous reference on deep-equality (`liveTerminalOrder.ts:70-80`) — the #159 update-depth fix, paired with the `sameNotification` guard in `KernelNotifications.tsx`.

## Requirement verification
- **Stable cards/order**: fixed 208×152 in all lifecycle states, creation-order append, slot-preserving reconcile, tree-canonical panes surviving catalog-join lag; settled `TerminalPreview` does one read and never polls (`live === false`).
- **Route-only authorization / no internal ids**: coordinator capability + operationId replay hash (`realtimeCoordinator.ts:350-389`); spoken names/task titles only *select*, never authorize; `boundRealtimePublicContext` bounds 64/validates routes/strips unsafe titles; `normalizeAgentName` rejects `pp_*`; `useAttentionRows` id-prefix fallback removed.
- **Realtime app control (#154)**: `realtime.app.request` is shape-bounded by the contract parser, then executes only against a fixed 9-entry benign destination allowlist and exactly-matching *visible registered* controls on the current screen (`realtimeAppControl.ts:3-13,60-74`); the only registered controls are five Settings navigation buttons. Worst case under a rogue provider is navigation — no mutation path.
- **Reconnect stability (#155)**: duplicate-socket guard, retire/reject-pending/backoff, stale host-frame fence (`muxrClient.ts:99-188,245-260`); real relay/host/client load test incl. five-agent route-loss flow; emulator exercised backgrounding, route loss, stack-restart reconnect, force-stop restart with no staleness.
- **Contract/crypto boundary (#157)**: deleted control-plane infrastructure has zero remaining importers; moved schemas live mobile-side (`typesRaw.ts`/`apiTypes.ts`); @muxr/crypto grant creation/verification, peer allowlist, broker capability checks, and inbound `peerGrantAllowsRequest` all intact.
- **Codex voice (#158)**: subscription credential stays host-side (owner-only `~/.codex/auth.json`, refreshed via local app-server, used only in the Authorization header — never logged, bundled, or sent to the client; no literals in plugin manifests). Foreground mic ordering is correct per AGENTS.md: `startVoiceService` → poll `voiceForegroundReady` (set only after `startForeground` with MICROPHONE type, `VoiceOverlayService.kt:480-490`) → `routeVoiceAudio` → `getUserMedia` (`realtimeWebRtc.ts:50-55,173-178`). Native streaming speech-to-speech preserved; provider policy stays in backend plugins.
- **#159 emulator fixes**: reference-stable reconcile + notification dedup; the API 36 gate reproduced neither prior failure across six churn scenarios and confirmed no lifecycle/title/name/provider staleness.
- **No secrets** in the full diff (credential-pattern scan clean; the only regex hits are `<task-notification>` test fixtures). **No legacy terminology** anywhere.

## Evidence
- CI on this exact SHA: `ci`/`suite` green — 32/32 passed, 2 skipped (no herdr server); CodeQL green. PR is MERGEABLE (state: BEHIND main — a rebase onto `bb937a44` before merge would be tidy; no conflicts, this branch does not touch #152's `agentKinds` region).
- Integration pane `final-integration-omp` (w18M:p1), pinned to this exact SHA: `yarn run check` **34/34 passed in 141.9s**; both typechecks; layout/contract/pane-titler/plugin self-checks; debug+release x86_64 builds; API 36 (`muxr_sandbox`) verification of both prior emulator failures; post-#159 Codex WebRTC smoke (onAddTrack → AudioTrack playout → stopPlayout) and Gemini PCM path/teardown; environment restored, temp Codex PATH shim removed.
- Locally re-ran at HEAD: 21/21 across `liveTerminalOrder`/`herd`/`herdTree`/`sessionSync.integration`/`realtimeSession`/`realtimeWebRtc` specs plus `providerRefusal.spec.mjs` (8/8).

## Non-blocking notes (no action required for merge)
1. `realtimeWebRtc.ts:82` passes no `iceServers`; live API 36 audio proves peer-reflexive ICE works against the provider's public candidates, but a STUN fallback would harden rare symmetric-NAT cases. Related: `acceptAnswer` (122-125) trusts the signaling channel for peer identity — standard WebRTC model, worth a comment; and `realtimeWebRtc.spec.ts` covers only the happy path.
2. `realtimeSession.ts:309-327` has no per-`requestId` dedup for app requests; actions are idempotent navigation so impact is nil — cheap fence anyway.
3. Pre-existing (unchanged by this PR, verified identical at base `a8614903`): the `openedSessions` reconnect-resync registry in `sync.ts:263,510-541` is only populated by its own private path and is effectively dormant — worth a follow-up issue; and `waitUntilClientOpen` (sync.ts:57-67) leaks its state listener on timeout.
4. Coordinator `resolve()` spans all agents on the machine rather than scoping to the capability's cwd — pre-existing design (names select, capability authorizes); consider scoping later.
5. Provider transcript text renders verbatim in `RealtimeConversation` — model prose over sanitized inputs, so acceptable; a display-side internal-id scrub would be cheap defense-in-depth.

Approved for merge at `4482bed7`.

PR #153, opened 2026-08-28
Make Herdr Agent Names canonical

umeranjum17:
## Stack

Stacked on #151. Base: `feat/compact-agent-activity`. Merge #151 first; do not merge this PR independently.

## Summary

- Treat Herdr `agent.name` as the canonical user-editable Agent Name while keeping Agent Route as the only authorization key and `displayName` only for wire compatibility.
- Move Agent Name defaults and editing into `plugins/pane-titler`: preserve existing public Herdr names, hide internal `pp_`/`pph_` names, assign an available deterministic animal without numeric suffixes, and expose native muxr read/rename RPC UI backed by `herdr agent rename`.
- Remove kernel Human Name allocation/reservation/duplicate policy and mobile new-agent naming inputs while keeping Task Title and Provider Kind separate.
- Replace the owned Human Name terminology and parser/export symbols with Agent Name across contract, host, mobile catalog/watch/spawn, architecture/use-case docs, and muxr skills. Excluded owner lanes remain untouched.

## Validation

- `yarn typecheck`
- `node apps/host/dist/agent/infrastructure/layoutSelfCheck.js`
- `node packages/contract/dist/selfCheck.js`
- `node plugins/pane-titler/agent-name.check.mjs`
- `node scripts/cli.mjs plugin check plugins/pane-titler`
- `node scripts/diagnostics/application/checkBundledPlugins.mjs`
- `npx vitest run apps/mobile/sources/catalog/application/sessionSync.integration.spec.ts --config apps/mobile/vitest.config.ts`
- `yarn check:architecture`
- `git diff --check`

## Known pre-existing risk

Standalone `yarn workspace @muxr/mobile typecheck` still reports two pre-existing errors at `apps/mobile/sources/terminal/presentation/TerminalView.tsx:222`: `TerminalViewProps` does not expose `onScroll`, and the resulting `nativeEvent` binding is implicitly `any`. This lane does not touch that prohibited owner file; all naming-lane TypeScript errors are cleared and the root typecheck passes.

PR #154, opened 2026-08-28
Add safe realtime agent and app control

umeranjum17:
## Stack

Stacked on #151. Base: `feat/compact-agent-activity`. Do not merge before the base PR.

## Summary

- add one bounded `send_agent_keybinding` tool that resolves Agent Name, Task Title, or uniquely resolved Provider Kind through the existing ambiguity-safe Agent Route path
- allow only the existing safe non-text Escape keybinding; unknown keys and ambiguous or unknown agents clarify without mutation
- reuse the authoritative session `sendKeys` path rather than maintaining a parallel interrupt abstraction, and never accept or expose pane IDs or routes in speech
- keep compact semantic mobile screen/control inspection separate from terminal key control, with allowlisted navigation and exact activation of visible registered controls without screenshots, coordinates, hidden routes, paths, prompts, terminal/file content, IDs, or credentials
- suffix only prompts delivered through muxr's realtime-agent transport with `came from a real-time agent`; ordinary chat and delegated prompt paths remain unchanged
- keep Agent Name allocation backend-owned during realtime start
- retain native streaming speech-to-speech and foreground-service microphone ordering

## Validation

- `yarn typecheck`
- `yarn workspace @muxr/mobile typecheck`
- `node packages/contract/dist/selfCheck.js`
- `npx vitest run apps/mobile/sources/conversation/application/realtimeSession.spec.ts --config apps/mobile/vitest.config.ts` — 6 passed
- `npx vitest run plugins/voice/providerRefusal.spec.mjs` — 8 passed
- `node apps/host/dist/agent/infrastructure/layoutSelfCheck.js`
- `yarn check:architecture`
- owned realtime/voice legacy terminology grep — empty
- `git diff --check`

## Residual adapter coverage

Semantic app-control tools are enabled on the bundled xAI adapter. Legacy OpenAI and Gemini adapters retain coding-agent and bounded keybinding tools but do not expose the new app-control tools. Unsupported or unregistered mobile UI safely clarifies without mutation or visual capture.

### DIFF
--- a/apps/host/src/agent/infrastructure/herdrSessionSource.ts
+++ b/apps/host/src/agent/infrastructure/herdrSessionSource.ts
@@ -159,6 +159,16 @@ interface AgentRecord {
     terminal_title?: string;
     terminal_title_stripped?: string;
 }
+export async function sendKeysToLiveAgent(
+    client: Pick<HerdrClient, 'call'>,
+    agentsByPane: ReadonlyMap<string, unknown>,
+    record: Pick<AgentIdentity, 'paneId' | 'displayName'>,
+    keys: string[],
+): Promise<void> {
+    if (!agentsByPane.has(record.paneId)) throw new Error(`${record.displayName} is not ready for keys.`);
+    await client.call('agent.send_keys', { target: record.paneId, keys });
+}
+
 
 interface PaneRecord {
     pane_id: string;
@@ -247,6 +257,7 @@ export async function createHerdrSessionSource(
             list: listRealtimeAgents,
             activity: async () => options.lifecycle?.catalog().events ?? [],
             start: startRealtimeAgent,
+            sendKeys: sendSessionKeys,
             prompt: promptSession,
             read: readSessionOutput,
             status: async (sessionId) => statusFor(sessionId).agentStatus ?? 'unknown',
@@ -1043,13 +1054,11 @@ export async function createHerdrSessionSource(
 
     async function startRealtimeAgent(input: {
         cwd: string;
-        displayName: string;
         taskTitle: string;
         kind: string;
     }): Promise<{ accepted: boolean; agent?: RealtimeCodingAgent }> {
         const result = await startSession({
             cwd: input.cwd,
-            displayName: input.displayName,
             taskTitle: input.taskTitle,
             label: input.taskTitle,
             kind: input.kind,
@@ -1068,6 +1077,11 @@ export async function createHerdrSessionSource(
         }
         await client.call('agent.prompt', { target: record.paneId, text });
     }
+    async function sendSessionKeys(sessionId: string, keys: string[]): Promise<void> {
+        const record = await resolvePane(sessionId);
+        await sendKeysToLiveAgent(client, agentsByPane, record, keys);
+    }
+
 
     async function readSessionOutput(sessionId: string): Promise<{ text: string; truncated: boolean }> {
         const record = await resolvePane(sessionId);
@@ -1908,10 +1922,7 @@ export async function createHerdrSessionSource(
                 .catch(() => undefined);
         },
 
-        async sendKeys(sessionId: string, keys: string[]): Promise<void> {
-            const record = await resolvePane(sessionId);
-            await client.call('agent.send_keys', { target: record.paneId, keys });
-        },
+        sendKeys: sendSessionKeys,
 
         async paneZoom(zoomOptions: {
             sessionId: string;
--- a/apps/mobile/sources/conversation/application/realtimeSession.ts
+++ b/apps/mobile/sources/conversation/application/realtimeSession.ts
@@ -6,6 +6,7 @@ import {
 import { acquireRealtimeCapture, type RealtimeCaptureLease } from './vadStandby';
 import { createRealtimePlayback } from '@/playback';
 import { sync } from '@/catalog/sync';
+import { realtimeAppController } from './realtimeAppControl';
 
 export type RealtimeStatus = 'connecting' | 'connected' | 'thinking' | 'speaking' | 'disconnected';
 
@@ -240,6 +241,26 @@ export function startRealtimeSession(options: {
                         return;
                     case 'realtime.transcript':
                         onTurn(frame.role, frame.text);
+                        return;
+                    case 'realtime.app.request':
+                        void (async () => {
+                            let ok = true;
+                            let text: string;
+                            try {
+                                if (frame.action === 'view') text = realtimeAppController.inspect();
+                                else if (frame.action === 'navigate') text = await realtimeAppController.navigateTo(frame.target!);
+                                else text = await realtimeAppController.activate(frame.target!);
+                            } catch {
+                                ok = false;
+                                text = 'The app could not complete that semantic action.';
+                            }
+                            if (!stopped && stream === next) next.send({
+                                type: 'realtime.app.result',
+                                requestId: frame.requestId,
+                                ok,
+                                text,
+                            });
+                        })();
                         return;
                     default:
                         return;
--- a/plugins/voice/stream.mjs
+++ b/plugins/voice/stream.mjs
@@ -8,12 +8,15 @@
  * only ever see the generic frame vocabulary.
  */
 import WebSocket from 'ws';
+import { randomUUID } from 'node:crypto';
 import { lstat, readFile } from 'node:fs/promises';
 import { homedir } from 'node:os';
 import { join } from 'node:path';
 import { createInterface } from 'node:readline';
 import { fileURLToPath } from 'node:url';
 import {
+    appTools,
+    appControlInstructions,
     cleanProviderProse,
     codingTools,
     isExplicitHangup,
@@ -29,7 +32,7 @@ const PROVIDER_URL = process.env.NODE_ENV === 'test' && process.env.MUXR_TEST_XA
 const root = process.env.MUXR_HOME?.trim() || join(homedir(), '.muxr');
 const keyFile = join(root, 'xai.key');
 let endAfterResponse = false;
-export const providerTools = codingTools;
+export const providerTools = [...codingTools, ...appTools];
 
 const MAX_STDOUT_BYTES = 256 * 1024;
 const MAX_REFUSAL_BODY_BYTES = 16 * 1024;
@@ -92,8 +95,8 @@ const state = (value, detail) => emit(detail === undefined
 const PROMPT = `You are the voice interface to a herd of coding agents. You are direct and brief. Speak with bright, upbeat energy and brisk enthusiasm; sound alert and helpful, never sultry or sleepy.
 
 <important>
-- Answer in one short sentence unless asked to elaborate. The user understands this work better than you do.
 ${voiceCoordinationInstructions}
+${appControlInstructions}
 - You do not do the work. The coding agent does. You carry instructions to it and report back what it did.
 - Assume the user is thinking out loud until they clearly ask for something.
 - Let them finish. A pause is thinking, not an invitation to speak: wait through it rather than filling it.
@@ -122,12 +125,41 @@ async function readKey() {
     return value;
 }
 
+const pendingAppRequests = new Map();
+const requestApp = (action, target) => {
+    if (action !== 'view' && (target === '' || Buffer.byteLength(target, 'utf8') > 160)) {
+        return Promise.resolve(action === 'navigate'
+            ? 'I could not find one app destination with that name. Ask me to inspect the app.'
+            : 'I could not find one visible control with that name. Ask me to inspect the app.');
+    }
+    return new Promise((resolve) => {
+        const requestId = randomUUID();
+        const timer = setTimeout(() => {
+            pendingAppRequests.delete(requestId);
+            resolve('The app did not answer that semantic request. Please try again.');
+        }, 15_000);
+        pendingAppRequests.set(requestId, {
+            resolve: (value) => {
+                clearTimeout(timer);
+                resolve(value);
+            },
+        });
+        if (!emit({ type: 'realtime.app.request', requestId, action, ...(target ? { target } : {}) })) {
+            clearTimeout(timer);
+            pendingAppRequests.delete(requestId);
+            resolve('The app could not receive that semantic request.');
+        }
+    });
+};
+
 let closing = false;
 const close = (reason, forceExit = false) => {
     if (closing) return;
     closing = true;
     clearTimeout(reconnectTimer);
     clearTimeout(stableTimer);
+    for (const pending of pendingAppRequests.values()) pending.resolve('The app semantic request was cancelled.');
+    pendingAppRequests.clear();
     clearTimeout(inputPoll);
     const timer = setTimeout(() => process.exit(forceExit ? 1 : 0), 1_000);
     emit({ type: 'realtime.closed', reason }, () => {
@@ -170,7 +202,12 @@ export function providerError(error) {
     const detail = cleanProviderProse(raw, 'provider error', 200);
     return { detail, terminal: /api key|auth|credit|quota|billing|permission|forbidden|invalid json|invalid payload|unknown name|unsupported/i.test(detail) };
 }
-const runTool = (name, input, operationId) => runCodingTool(name, input, operationId);
+const runTool = (name, input, operationId) => {
+    if (name === 'inspect_app') return requestApp('view');
+    if (name === 'navigate_app') return requestApp('navigate', text(input?.destination));
+    if (name === 'activate_app_control') return requestApp('activate', text(input?.control));
+    return runCodingTool(name, input, operationId);
+};
 
 const currentProvider = (current, epoch) => !stopped && ws === current && providerEpoch === epoch;
 const finishGracefulEndIfDrained = () => {
@@ -358,6 +395,14 @@ const clientFrameBytes = (frame) => frame.type === 'realtime.audio'
     : Buffer.byteLength(JSON.stringify(frame));
 
 function handleClientFrame(frame) {
+    if (frame.type === 'realtime.app.result') {
+        const pending = pendingAppRequests.get(frame.requestId);
+        if (pending !== undefined) {
+            pendingAppRequests.delete(frame.requestId);
+            pending.resolve(frame.text);
+        }
+        return;
+    }
     if (frame.type === 'realtime.control') {
         if (frame.action === 'stop') {
             stopped = true;
--- a/apps/host/src/agent/infrastructure/layoutSelfCheck.ts
+++ b/apps/host/src/agent/infrastructure/layoutSelfCheck.ts
@@ -13,6 +13,7 @@ import {
 import { lifecycleReasonForObservation } from '../domain/lifecycle.js';
 import { IdentityStore, parseTaskTitle } from './identity.js';
 import { RealtimeCodingCoordinator } from './realtimeCoordinator.js';
+import { sendKeysToLiveAgent } from './herdrSessionSource.js';
 import { createConnection } from 'node:net';
 import { mkdtempSync, writeFileSync } from 'node:fs';
 import { tmpdir } from 'node:os';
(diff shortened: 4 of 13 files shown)


umeranjum17 (commented):
**Verdict: REQUEST CHANGES** *(GitHub blocks the formal request-changes state on the author's own PR — HTTP 422 "Can not request changes on your own pull request" — so this review is submitted as a comment; treat it as request-changes.)*

Strict read-only review of the diff vs base `feat/compact-agent-activity`, checked against repository standards (CONTEXT.md / CONTRIBUTING.md / AGENTS.md test policy) and the stated requirements. All focused checks were re-run at PR head (`f765b54`): `plugins/voice/providerRefusal.spec.mjs` (8 passed), `apps/mobile/.../realtimeSession.spec.ts` (6 passed), contract `selfCheck` and `layoutSelfCheck` (both pass from source).

## Verified against requirements

- **Native streaming speech-to-speech retained** — `plugins/voice/stream.mjs` remains the xAI realtime WS speech-to-speech adapter; no STT+LLM+TTS pipeline; mic foreground-service ordering untouched.
- **One bounded keybinding tool via the authoritative route and existing sendKeys path** — `send_agent_keybinding` is the only key tool; the allowlist check runs **before** target resolution (`realtimeCoordinator.ts:390-391`), the literal `['escape']` is sent rather than provider text (`:394`), it is replay-fenced like `prompt`, and it flows through the same `resolve()` and `SessionSource.sendKeys` → `agent.send_keys` used everywhere else. No parallel interrupt abstraction exists.
- **sendKeys unification** — `sendSessionKeys` now targets `record.agentName` (matching `agent.prompt`/`agent.wait`) instead of `record.paneId`. Sole other caller is `session.answer` → `answerAgent` (blocked-agent y/n), and a blocked agent always has a live herdr agent name; the not-ready throw surfaces as the coordinator's generic error, never raw text.
- **Ambiguity-safe resolution** — Agent Name / Task Title / unique Provider Kind resolve through `aliases()`; >1 match returns a clarification with no mutation (proved by `layoutSelfCheck` `key-ambiguous` and the spec's duplicate-Maria case). Unknown key clarifies with zero `sendKeys` calls.
- **No pane IDs/routes from speech** — spoken targets match only displayName/kind/taskTitle aliases; sessionIds are never aliased; clarifications pass through `safeProviderText` which redacts id/path-like content; handler exceptions collapse to a generic line (`realtimeCoordinator.ts:462-463`).
- **Provenance suffix scoped correctly** — `\n\ncame from a real-time agent` is appended only in the coordinator's `prompt` handler (`realtimeCoordinator.ts:401`); a coordinator capability is issued only when `voiceSession` is true (`herdrSessionSource.ts:1424`), so only realtime-transport prompts are stamped. Ordinary `SessionSource.prompt` (`herdrSessionSource.ts:1977`) and delegated paths are untouched. Exactness asserted in both self-check and spec.
- **App inspection privacy** — `inspect()` exposes only a semantic screen name, registered control labels, and static destination labels; navigation is allowlist-only (raw path `/settings` is refused in the test); activation requires exactly one visible registered control; `realtime.app.request`/`.result` frames are contract-validated at both the host (`pluginStreamManager.ts:328,390`) and mobile (`openPluginStream.ts:241,325`) boundaries. All nine `DESTINATIONS` paths exist as expo-router routes.
- **Backend-owned Agent Name at realtime start** — `name` removed from the tool schema and coordinator; `identity.reserve(undefined)` allocates from the pool.

## Blocking

1. **A degenerate provider tool call tears down the live voice session** (`plugins/voice/stream.mjs:199-201`, `:141`). `navigate_app`/`activate_app_control` forward the provider-supplied string into `realtime.app.request` with only `text()` trimming. Two provider-controlled inputs produce a frame the host rejects, and `pluginStreamManager.ts:340-343` responds to any invalid host frame with `attachment.close('plugin stream emitted an invalid realtime frame')` — killing the whole realtime session:
   - empty/whitespace `destination` → `target` falsy → omitted → `parseRealtimeHostFrame` throws (`action !== 'view'` with no target);
   - any target > 160 UTF-8 bytes (likely when the model echoes a long spoken phrase) → `boundedText(frame.target, 160, ...)` throws.

   Reproduced against `parseRealtimeHostFrame` at PR head: both cases throw `invalid realtime app target`. LLM tool arguments are untrusted input; the existing spec only sends well-formed targets so this path is uncovered. Fix is one guard in `runTool`/`requestApp`: when the target is empty or exceeds 160 bytes, resolve with the standard clarification ("I could not find one app destination with that name…") without emitting. No privacy or mutation impact — availability only — but it takes down the exact surface this PR ships.

## Non-blocking

2. **Residual `Human Name` in an owned realtime path** — `packages/contract/src/realtime/application/boundRealtimePublicContext.ts:21` still reads "Human Name never does". This contradicts the requirement (no Human Name/humanName vocabulary in owned paths) and the PR's own validation claim ("owned realtime/voice legacy terminology grep — empty").
3. **Glossary divergence** — `CONTEXT.md:41-43` still defines **Human Name** and explicitly lists "agent name" under *Avoid*, while `CONTRIBUTING.md:85-88` mandates that code names match CONTEXT.md. The rename to "Agent Name" is required here, but the committed glossary now asserts the opposite; the glossary should be updated somewhere in this stack so ubiquitous language stays single-sourced.

Requesting changes for finding 1; everything else in the diff holds up well under the authorization/privacy/provenance lens.

plugins/voice/stream.mjs
```diff
@@ -170,7 +195,12 @@ export function providerError(error) {
     const detail = cleanProviderProse(raw, 'provider error', 200);
     return { detail, terminal: /api key|auth|credit|quota|billing|permission|forbidden|invalid json|invalid payload|unknown name|unsupported/i.test(detail) };
 }
-const runTool = (name, input, operationId) => runCodingTool(name, input, operationId);
+const runTool = (name, input, operationId) => {
+    if (name === 'inspect_app') return requestApp('view');
+    if (name === 'navigate_app') return requestApp('navigate', text(input?.destination));
+    if (name === 'activate_app_control') return requestApp('activate', text(input?.control));
```
umeranjum17, line 208: **Blocking — provider-controlled input can terminate the live voice session.**

`text(input?.destination)` / `text(input?.control)` flow unvalidated into `requestApp`, which emits `realtime.app.request` (line 141). Two ordinary degenerate LLM tool calls produce a frame the contract rejects:

- empty/whitespace destination → `target` falsy → omitted → `parseRealtimeHostFrame` throws (`action !== 'view'` requires a target);
- any destination/control over 160 UTF-8 bytes → `boundedText(frame.target, 160, 'app target')` throws.

The host treats any invalid plugin host frame as fatal (`pluginStreamManager.ts:340-343`: `attachment.close('plugin stream emitted an invalid realtime frame')`), so one malformed tool argument tears down the user's entire realtime session. Reproduced both throws against `parseRealtimeHostFrame` at PR head.

Suggested fix: guard here (or in `requestApp`) — if the target is empty or exceeds 160 bytes, return the standard clarification string to the provider without emitting a frame. The existing spec only exercises well-formed targets, so this path is currently uncovered.

apps/mobile/sources/conversation/application/realtimeSession.ts
```diff
@@ -241,6 +242,26 @@ export function startRealtimeSession(options: {
...
+                            let text: string;
+                            try {
+                                if (frame.action === 'view') text = realtimeAppController.inspect();
+                                else if (frame.action === 'navigate') text = await realtimeAppController.navigateTo(frame.target!);
+                                else text = await realtimeAppController.activate(frame.target!);
+                            } catch {
+                                ok = false;
+                                text = 'The app could not complete that semantic action.';
+                            }
+                            if (!stopped && stream === next) next.send({
```
umeranjum17, line 257: Nit: `next.send` sits outside the `try`/`catch`; `openPluginStream`'s send validates via `parseRealtimeClientFrame` and can throw, which would surface as an unhandled rejection from this `void` IIFE. Practical risk is low (controller texts are short and bounded), but wrapping the send or catching on the IIFE would make the handler total.

umeranjum17 (commented):
**Verdict: APPROVE** — reviewed HEAD `fea2cfd52d7ac816ca49193ec116dc4a738895bb`. *(GitHub's API blocks the formal approve state on the author's own PR, so this is submitted as a comment review; treat it as an approval.)*

Follow-up to [review 5051585021](https://github.com/umeranjum17/muxr/pull/154#pullrequestreview-5051585021). Delta vs prior head `f765b549` touches exactly 7 files: the `requestApp` guard, spec hardening, and the `parseHumanName` → `parseAgentName` rename. All prior blockers are resolved; no blocking findings remain.

## Prior blocker — resolved

**Degenerate provider targets no longer kill the voice session.** `plugins/voice/stream.mjs:130-134` now rejects `navigate`/`activate` targets that are empty (`text()` trims, so whitespace-only collapses to `''`) or exceed 160 UTF-8 bytes, resolving the tool call with the standard clarification **before any frame is emitted** — no `realtime.app.request`, no mutation, no host-side parse failure, no `attachment.close`. The bound exactly matches the contract parser (160 bytes allowed, >160 rejected; empty rejected on both sides).

**Focused assertions now cross the real parser boundary.** `providerRefusal.spec.mjs` parses every child stdout line through the actual `parseRealtimeHostFrame` and asserts `invalidHostFrames === 0`; degenerate calls (`destination: ' \t '`, `control: 'é'.repeat(81)` = 162 UTF-8 bytes) assert the exact clarification strings, zero new `realtime.app.request` frames, and a subsequent `agent_status` call proves the stream is still alive and functional. Re-ran at HEAD: **8/8 passed**.

## Prior non-blocking findings

- **Residual `Human Name` in owned realtime paths — resolved.** `boundRealtimePublicContext.ts:21` now reads "Agent Name never does", and the underlying `parseHumanName`/`HUMAN_NAME` were renamed to `parseAgentName`/`AGENT_NAME` as a clean cutover: no shim, no dangling callers (repo-wide grep empty), contract `tsc --noEmit` clean, old error-string dependents none. Case-insensitive sweep of `plugins/voice*`, `packages/contract/src/realtime`, `apps/mobile/sources/conversation`, and the host realtime infrastructure is clean. (`agent/infrastructure/identity.ts` and herd-domain `lifecycleEventHumanName` still use the term, but those are the shared agent-identity store outside the realtime-owned surface and consistent with the committed glossary.)
- **Glossary divergence — still open, still non-blocking.** `CONTEXT.md:41-43` continues to define **Human Name** with *Avoid: agent name* while owned paths now standardize on Agent Name; worth reconciling somewhere in this stack.
- **Nit at `realtimeSession.ts:257`** (`next.send` outside try/catch) — unchanged, remains a nit.

## Re-verified requirement checks at HEAD

- **Bounded keybinding design** — `realtimeCoordinator.ts`, `herdrSessionSource.ts`, and `coordinatorPolicy.mjs` are byte-identical between heads. Single `send_agent_keybinding` tool with `enum: ['escape']`; coordinator allowlist check precedes resolution; the literal `['escape']` is sent through the existing `SessionSource.sendKeys` → `agent.send_keys` path; no parallel interrupt abstraction; no mutation on ambiguity or unknown key; ids/routes never spoken (aliases cover only Agent Name / Task Title / Provider Kind; clarifications pass `safeProviderText` redaction; handler failures collapse to a generic line).
- **Provenance** — `\n\ncame from a real-time agent` appended only in the coordinator prompt handler; coordinator capabilities issued only for `voiceSession` streams; ordinary `SessionSource.prompt` and delegated paths unchanged. Exactness asserted in `layoutSelfCheck` and the voice spec.
- **No new regressions** — the rename is mechanical with full caller migration; the guard is purely additive; native streaming speech-to-speech and mic foreground-service ordering untouched.

## Verification at `fea2cfd5`

- `plugins/voice/providerRefusal.spec.mjs` — 8/8 passed (incl. new degenerate-target + real-parser-boundary assertions)
- `apps/mobile/.../realtimeSession.spec.ts` — 6/6 passed
- contract `selfCheck` — PASS; host `layoutSelfCheck` — PASS (both from source)
- `tsc --noEmit -p packages/contract` — clean
- CI: `suite` check **completed/success** on this commit

umeranjum17 (commented):
**Verdict: REQUEST CHANGES** — reviewed HEAD `6a2220a18de0eee3a8cd5c5b2258a1a774a18bee` against base `d47529ee5a5e857abf628c36bd9c87777e548e33`. *(GitHub's API blocks the formal request-changes state on the author's own PR, so this is a comment review; treat it as request-changes.)*

Follow-up to [5051585021](https://github.com/umeranjum17/muxr/pull/154#pullrequestreview-5051585021) and [5051751475](https://github.com/umeranjum17/muxr/pull/154#pullrequestreview-5051751475). The rebase onto merged #153 is clean everywhere except one function, where the conflict resolution did not follow #153's canon.

## Blocker

1. **`sendSessionKeys` routes a mutation by a runtime-mutable Agent Name** (`apps/host/src/agent/infrastructure/herdrSessionSource.ts:1070-1073`, inline comment). Merged #153 makes herdr agent names pane-titler-owned display labels — `plugins/pane-titler/agent-name.mjs` renames agents at runtime (`agent rename` in `ensureAgentName`/`renameAgent`) — and accordingly migrated every sibling mutation to pane-id targets: `agent.prompt` targets `record.paneId` with `agentsByPane.has(paneId)` readiness (`:1063-1068`), and the interrupt path targets `record.paneId` with a `pane.send_keys` fallback (`:1953`). The PR's `sendSessionKeys` still targets `record.agentName` with the old `agentName === undefined` readiness check. Consequences:
   - voice Escape and blocked-agent y/n answers (`session.answer` → `answerAgent` → `sendKeys`) can miss or fail whenever pane-titler renames the agent between `resolvePane`'s snapshot refresh and the `agent.send_keys` call — an unavoidable TOCTOU when routing by name;
   - it **regresses `session.answer`** from the base's pane-id routing to display-name routing;
   - it violates the invariant this re-review was chartered to confirm and that the repo documents everywhere (CONTEXT.md, host README agent domain, #153 itself): *Agent Name is display-only, never a routing key*;
   - the coordinator specs fake `handlers.sendKeys` above the herdr-targeting boundary, so no test catches it.

   Fix is two lines — mirror `promptSession`: readiness via `agentsByPane.has(record.paneId)`, target `record.paneId`.

## Conflict resolution otherwise verified

- **Canonical Agent Name implementation preserved, no shim/duplicate parser.** Exactly one `parseAgentName`/`normalizeAgentName` pair exists (`packages/contract/src/herd/domain/sessionState.ts:188-196`, from #153, now also hiding internal `pp_*` names behind `'Agent'`); `boundRealtimePublicContext` consumes it. Repo-wide sweep for `parseHumanName`/`HUMAN_NAME`/`nextHumanName`/`isValidHumanName`/legacy vocabulary in contract, host agent, voice plugins, and mobile conversation paths: **empty**. The PR's contract delta from the previous head dissolved into the base as expected — the PR diff is back to 13 files touching none of the renamed contract surfaces.
- **All previously approved realtime requirements carry over byte-identically** — `realtimeCoordinator.ts`, `stream.mjs`, `coordinatorPolicy.mjs`, `providerRefusal.spec.mjs`, contract `realtimeStream.ts`, and every mobile realtime file are unchanged from approved head `fea2cfd5`: target boundary guard (empty/>160-byte rejected before emit, no stream close), spec assertions crossing the real `parseRealtimeHostFrame` with `invalidHostFrames === 0` and post-degenerate liveness, Escape-only enum with allowlist-before-resolve and literal `['escape']`, no mutation on ambiguity/unknown key, no ids/routes spoken, exact `came from a real-time agent` suffix only on voice-session coordinator prompts, native streaming speech-to-speech untouched.

## Verification at `6a2220a1`

- `plugins/voice/providerRefusal.spec.mjs` — 8/8 passed
- `apps/mobile/.../realtimeSession.spec.ts` — 6/6 passed
- contract `selfCheck` (source + built dist) — PASS; host `layoutSelfCheck` — PASS (against the head-built contract)
- `tsc --noEmit -p packages/contract` — clean
- CI: `suite` **completed/success** on this commit

Approval will follow once `sendSessionKeys` routes by pane id like its siblings; everything else is ready.

apps/host/src/agent/infrastructure/herdrSessionSource.ts
```diff
@@ -1068,6 +1067,12 @@ export async function createHerdrSessionSource(
         }
         await client.call('agent.prompt', { target: record.paneId, text });
     }
+    async function sendSessionKeys(sessionId: string, keys: string[]): Promise<void> {
+        const record = await resolvePane(sessionId);
+        if (record.agentName === undefined) throw new Error(`${record.displayName} is not ready for keys.`);
+        await client.call('agent.send_keys', { target: record.agentName, keys });
```
umeranjum17: **Blocking — conflict resolution kept the pre-#153 routing convention here.**

Merged #153 makes `agent.name` a pane-titler-owned, runtime-mutable display label (`plugins/pane-titler/agent-name.mjs` calls `agent rename` in both `ensureAgentName` and `renameAgent`) and migrated the sibling mutations to pane-id targets: `promptSession` above (`:1063-1068`) checks `agentsByPane.has(record.paneId)` and targets `record.paneId`, and the interrupt path at `:1953` targets `record.paneId` with a `pane.send_keys` fallback.

This function still gates on `record.agentName === undefined` and targets `record.agentName`, so:
- voice Escape and `session.answer` y/n keys route by a mutable name — a rename landing between `resolvePane`'s refresh and this call misses or fails;
- `session.answer` regresses from the base's pane-id routing;
- it contradicts the repo invariant that Agent Name is display-only and never a routing key.

Suggested fix, mirroring `promptSession`:
```ts
async function sendSessionKeys(sessionId: string, keys: string[]): Promise<void> {
    const record = await resolvePane(sessionId);
    if (!agentsByPane.has(record.paneId)) throw new Error(`${record.displayName} is not ready for keys.`);
    await client.call('agent.send_keys', { target: record.paneId, keys });
}
```

The coordinator specs fake `handlers.sendKeys` above this boundary, so this is untested — worth one assertion at the herdr-call level if cheap.

umeranjum17 (commented):
**Verdict: APPROVE** — reviewed HEAD `64a4ca7926dfe019c525f65f93a11e0e546b5c18` against base `d47529ee`. *(GitHub's API blocks the formal approve state on the author's own PR; treat this comment review as an approval.)*

Follow-up to [5051859189](https://github.com/umeranjum17/muxr/pull/154#pullrequestreview-5051859189). The delta vs the previously reviewed head `6a2220a1` touches exactly two files — `herdrSessionSource.ts` and `layoutSelfCheck.ts` — and resolves the sole blocker precisely as prescribed. Everything else is byte-identical to the already fully reviewed head. No blockers remain.

## Blocker resolution — verified

**`sendSessionKeys` now routes only by resolved pane id with live readiness.** The new exported `sendKeysToLiveAgent` (`herdrSessionSource.ts:162-170`) gates on `agentsByPane.has(record.paneId)` and calls `agent.send_keys` with `target: record.paneId` — mirroring `promptSession` exactly. `sendSessionKeys` (`:1080-1083`) resolves through `resolvePane` (fresh snapshot + pane-liveness check) and delegates. Both consumers ride this single path: the realtime coordinator's Escape handler (`sendKeys: sendSessionKeys`, `:260`) and `session.answer` → `answerAgent` via the public `SessionSource.sendKeys` (`:1925`). No `agentName` routing remains anywhere in the key path — the remaining `agentName` occurrences are identity/display metadata only, never targets.

**The herdr-boundary gap is closed.** `layoutSelfCheck.ts:166-176` now drives the real `sendKeysToLiveAgent` with a recording client immediately after the Agent Name reuse scenario and asserts the emitted call is exactly `agent.send_keys { target: 'w1:p2', keys: ['escape'] }` — proving runtime Agent Name reuse cannot redirect Escape or `session.answer` away from the authoritative pane. This is asserted below the coordinator's faked `handlers.sendKeys`, at the herdr call level, which was the missing coverage.

## Regressions

None. `git diff 6a2220a1..64a4ca79` is confined to the fix and its assertion; all realtime coordinator, voice plugin, contract, and mobile files are unchanged from the reviewed state (target boundary guard, real-parser spec assertions, Escape-only allowlist, ambiguity no-mutation, no ids spoken, exact realtime-only provenance suffix, native streaming voice — all carry over byte-identically).

## Verification at `64a4ca79`

- host `layoutSelfCheck` — PASS (against head-built `@muxr/contract` + `@muxr/crypto`), including the new herdr-boundary assertion
- contract `selfCheck` (built dist) — PASS
- `plugins/voice/providerRefusal.spec.mjs` — 8/8 passed
- `apps/mobile/.../realtimeSession.spec.ts` — 6/6 passed
- CI: `suite` **completed/success** on this commit

PR #155, opened 2026-08-28
fix: stabilize mobile reconnects under agent load

umeranjum17:
## Stack

Stacked on #151. Base: `feat/compact-agent-activity`. Do not merge before #151 is ready.

## Root causes

The original load flap came from mobile RPC timers surviving WebSocket loss. After a replacement socket received authenticated host traffic and became open, old 20-second timers demoted that healthy socket to stale; the next frame reopened it and triggered another catalog reconciliation. With five active agents, each `session.list` required a measured 103–116 ms Herdr snapshot, amplifying the stale/open cycle into repeated catalog batches.

A Tailscale off/on transition exposed a related but separate recovery gap: React Native could retain a half-open control or terminal socket whose `readyState` still said OPEN. Foreground terminal recovery treated OPEN as healthy and did not replace it. Ticket/grant acquisition can also spend up to 10 seconds in each awaited network leg, so closing or superseding a client during that await previously allowed a late socket/state resurrection.

Bounded live evidence for the load incident showed 232 `session.list` and 231 `machines.list` requests, bursts near five catalog cycles per second, no host request failures, no relay restart loop, empty TCP queues, and no descriptor/process pressure.

## Fix

- Re-check closed/socket ownership after awaited grant and ticket acquisition, before WebSocket creation/assignment.
- Ignore catch-path state and permanent-error callbacks after close or after another open claimed the socket.
- Parenthesize initial connection-ownership precedence and prevent concurrent-open orphan sockets.
- Clear and reject pending RPCs when their owning socket closes; cap pending RPCs at 128.
- Fence stale timeout decisions by current socket plus authenticated-host-frame revision.
- Immediately retire a half-open current socket and schedule a fresh ticketed connection when a request times out without newer authenticated host traffic.
- Rebuild the control connection when the native app returns to foreground, ensuring a network-route change obtains fresh grant/ticket reachability.
- Force a terminal reattach/full repaint on foreground instead of accepting a nominally OPEN stale terminal socket.
- Guard terminal socket creation after awaited ticket acquisition.
- Preserve cached agent lifecycle truth across transient transport loss.

E2EE, replay protection, relay routing authority, and native voice paths remain unchanged.

## Regression flows

The existing real client → relay → host WebSocket flow now covers:

- Five active sessions with measured 105 ms catalog latency.
- Relay interruption during an in-flight sync without a stale/open flap.
- Five concurrent catalog reads and health through the relay's first 30-second server ping.
- A route-loss transition where the relay remains up but the machine disappears: the unanswered request must produce `stale → closed → connecting`, retire the half-open socket, and recover five-session sync when the host returns.
- Close during delayed ticket acquisition: no WebSocket creation or state resurrection.
- Close during delayed grant failure: stable closed state and no permanent-error callback.

## Validation

- Focused mobile control/terminal recovery: 6 tests passed.
- Host connection flows: 7 tests passed.
- Relay routing/persistence flows: 6 tests passed.
- Real five-agent WebSocket reconnect/keepalive suite: 2 tests passed in 33.6 seconds.
- Mobile typecheck passed.
- Root TypeScript build passed.
- Mobile/host/relay architecture tests and both architecture scripts passed.
- `git diff --check` passed.

## Incident impact and residual risks

The currently installed mobile binary does not contain PR #155. It still has the old pending-timer amplification and lacks foreground/half-open socket replacement, so it explains both the slow recovery and why the phone can remain wedged after the host is reachable again. Desktop access remaining healthy is consistent with a mobile-path recovery defect rather than host failure.

Physical radio handoff, Tailscale daemon behavior, WAN loss, and proxy buffering were not reproduced on-device. The flow uses real local WebSockets in development cleartext mode; existing encrypted peer flows cover E2EE behavior. `session.list` still costs roughly 105 ms under this load but is stable after request amplification is removed.

### DIFF
--- a/apps/mobile/sources/app/_layout.tsx
+++ b/apps/mobile/sources/app/_layout.tsx
@@ -22,7 +22,7 @@ import { SidebarNavigator } from '@/herd/ui';
 import sodium from '@/encryption/libsodium.lib';
 import { View, Platform, AppState, Pressable, Text } from 'react-native';
 import { ModalProvider } from '@/modal';
-import { syncRestore } from '@/catalog/sync';
+import { syncReconnect, syncRestore } from '@/catalog/sync';
 import { FaviconPermissionIndicator } from '@/components/web/FaviconPermissionIndicator';
 import { CommandPaletteProvider } from '@/components/CommandPalette/CommandPaletteProvider';
 import { StatusBarProvider } from '@/components/StatusBarProvider';
@@ -313,6 +313,17 @@ export default function RootLayout() {
             }, 100);
         }
     }, [initState]);
+
+    React.useEffect(() => {
+        if (initState?.credentials === undefined || Platform.OS === 'web') return;
+        let previous = AppState.currentState;
+        const subscription = AppState.addEventListener('change', (next) => {
+            const resumed = next === 'active' && previous !== 'active';
+            previous = next;
+            if (resumed) void syncReconnect().catch(() => undefined);
+        });
+        return () => subscription.remove();
+    }, [initState?.credentials]);
 
     const handledNotificationIds = React.useRef<Set<string>>(new Set());
     const handleNotificationResponse = React.useCallback(async (response: Notifications.NotificationResponse | null) => {
--- a/apps/mobile/sources/pairing/infrastructure/muxrClient.ts
+++ b/apps/mobile/sources/pairing/infrastructure/muxrClient.ts
@@ -26,7 +26,7 @@ import {
 } from '@muxr/contract';
 import { DeviceV2Crypto, refreshHostedGrant, type StoredHostedGrant } from '../application/hostedE2ee';
 
-/** `stale`: socket is up but the host stopped answering requests (host process gone). */
+/** `stale`: host liveness is unproven because a request timed out without newer authenticated host traffic. */
 export type ConnectionState = 'connecting' | 'open' | 'closed' | 'stale';
 
 export interface MuxrClientOptions {
@@ -68,6 +68,7 @@ function requestFailure(type: RequestType, error: string, code?: string): MuxrRe
 type EventListener = (sessionId: string, event: SessionEvent) => void;
 type StateListener = (state: ConnectionState) => void;
 type PluginInvalidationListener = (frame: Extract<HostFrame, { type: 'plugins.invalidated' }>) => void;
+const MAX_PENDING_REQUESTS = 128;
 
 export class MuxrClient {
     private socket: WebSocket | undefined;
@@ -81,6 +82,8 @@ export class MuxrClient {
     private readonly clientId = nextRequestId('client');
     private reconnectTimer: ReturnType<typeof setTimeout> | undefined;
     private reconnectAttempt = 0;
+    /** Valid host traffic observed on the current socket. Request timers use it as a liveness fence. */
+    private hostFrameRevision = 0;
 
     state: ConnectionState = 'closed';
 
@@ -98,7 +101,12 @@ export class MuxrClient {
     }
 
     private async open(): Promise<void> {
-        if (this.closed) return;
+        if (this.closed || (this.socket !== undefined
+            && (this.socket.readyState === WebSocket.CONNECTING || this.socket.readyState === WebSocket.OPEN))) return;
+        if (this.reconnectTimer !== undefined) {
+            clearTimeout(this.reconnectTimer);
+            this.reconnectTimer = undefined;
+        }
         this.setState('connecting');
         const { machineId, token } = this.options;
         let relayUrl = this.options.relayUrl;
@@ -124,6 +132,7 @@ export class MuxrClient {
                     transport: 'relay',
                 }), 'relay');
         } catch (error) {
+            if (this.closed || this.socket !== undefined) return;
             const rejected = error instanceof WsTicketError && (error.status === 401 || error.status === 403);
             const expired = error instanceof Error && /grant expired/i.test(error.message);
             const permanent = expired || rejected && this.hosted?.grant.source === 'selfhost';
@@ -144,10 +153,15 @@ export class MuxrClient {
             }
             return;
         }
+        if (this.closed || this.socket !== undefined) return;
         const socket = new WebSocket(url);
         this.socket = socket;
 
         socket.onopen = () => {
+            if (this.socket !== socket) {
+                socket.close();
+                return;
+            }
             this.reconnectAttempt = 0;
             // The relay accepts a client peer even when no machine is attached,
             // so socket open is not "connected". Stay `connecting` until the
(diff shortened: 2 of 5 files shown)


umeranjum17:
## Review — verdict: **REQUEST CHANGES**

Reviewed HEAD: `61ca6180087371814c79730d7d28840599e58c61` against base `feat/compact-agent-activity` (merge-base `f31eb495`). GitHub blocks a formal review from the PR author's auth; this comment verdict is authoritative.

### Blocking

**B1 — `close()` racing an in-flight `open()` resurrects the client and rewrites closed state** (`apps/mobile/sources/pairing/infrastructure/muxrClient.ts`).

`open()` awaits `refreshHostedGrant` (hosted) / `issueWsTicket` (any ticketed connect) between the top guard (L104–105) and socket creation (L155–156), and never re-checks `this.closed` after those awaits:

- `close()` lands in that window → the in-flight `open()` still executes `new WebSocket(url); this.socket = socket` on a closed client. `onopen`'s identity fence passes (`this.socket === socket`), `client.hello` is sent, and the first host frame flips state to `open`. `onclose`'s `if (this.closed) return` only prevents *rescheduling*; the zombie socket itself lives indefinitely (relay keepalive sustains it).
- The failure path has the same hole: the `catch` (L134+) can `setState('stale')` and fire `onPermanentError` after `close()`.
- This is reachable, not theoretical: `sync.ts` `bootstrap()` (L559–560) and `reconnect()` (L688–689) do `client.close()` then build a **new** client, while the old client retries every 1.5–30 s and spends an HTTP roundtrip inside `open()` on each attempt. The old client's `onStateChange` listener stays attached, so the zombie rewrites `socketStatus` and triggers `refreshCatalog`/resyncs concurrently with the replacement client — dueling state writers, the exact flap class this PR exists to eliminate, and after a credential/machine switch the zombie reports "connected" for the *old* machine.

Pre-existing shape, but this PR's stated contract is "explicit close … stabilizes closed state" and "a superseded actor cannot rewrite healthy connection state," and it rewrites exactly these functions. One-line fix inside `open()` after the try/await block (and before the catch-path `setState`): bail when `this.closed || this.socket !== undefined`. That same guard also closes N1 below.

### Non-blocking

- **N1 — concurrent-open orphan (latent).** Two `open()`s in flight (both past the L104 guard while `this.socket` is undefined): if socket A reaches OPEN before B assigns `this.socket = socketB`, A is silently superseded with all handlers fenced and **nobody ever closes it** — an orphan live connection. Unreachable from today's callers (`sync.ts` calls `connect()` exactly once per client; the reconnect timer is single and cleared at `open()` entry), so not blocking — but the B1 guard fixes it for free.
- **N2 — stale-semantics narrowing.** With the `hostFrameRevision` fence, a host that streams `session.event`s but never answers RPCs is no longer demoted to `stale` (previously any timeout demoted). Defensible — transport is provably alive and each request still rejects — but it changes the `ConnectionState` doc-comment's meaning ("host stopped answering requests"); worth updating the comment.
- **N3 —** `this.closed || this.socket !== undefined && (…)` relies on `||`/`&&` precedence; correct, but parenthesize.

### Verified good

- **Pending cap = 128, correct error behavior.** `MAX_PENDING_REQUESTS = 128`; over-cap requests reject synchronously with an `Error` *before* any timer/map registration — no leak, no silent drop. Host-normalized `MuxrRequestError` path untouched. (Unrelated to contract's `MAX_RPC_PER_DEVICE=3` plugin caps — checked.)
- **Stale-timeout fencing.** Timers are triple-fenced (socket identity + `hostFrameRevision` + `state !== 'closed'`), and `onclose` now rejects+clears all pending for its socket, so stale timers from a lost socket cannot demote a healthy replacement. `hostFrameRevision` only increments in `handleMessage` after routing/E2EE validation succeeds — genuinely authenticated traffic in hosted mode.
- **Explicit close** (modulo B1): clears reconnect timer, rejects pending with timer cleanup, detaches `this.socket` *before* `socket.close()` so the fenced `onclose` cannot rewrite state or schedule reconnects; idempotent.
- **Lifecycle truth on reconnect.** Untouched by the diff; on re-open, `refreshCatalog` folds host truth (herd-tree `agentStatus` + `lifecycle.catalog`) and `mergeCatalogAgent`/`applyHostInfoToAgent` preserve cached `agentStatus`/`lifecycleStateSince`/`agentState` when the incoming DTO carries none. Nothing fabricates lifecycle client-side.
- **Regression flow is real and runs in CI.** Test drives real `ws` sockets through `startRelay` + `startHost` + `MuxrClient` with 5 sessions and injected 105 ms `session.list` latency; asserts no `stale` after relay restart, 5 concurrent catalog reads, and health across the relay's first server ping (confirmed `setInterval(… client.ping(), 30_000)` at `apps/relay/src/relay.ts:1169–1173`; test waits 30.5 s). It matches vitest's default include and demonstrably ran in CI: "unit: all vitest flows" went 14.7 s (base run) → 43.0 s (this HEAD), well inside the 150 s row timeout. The no-stale assertion would fail against the pre-fix code (old 500 ms timer demotes post-reopen).
- **No E2EE/replay/routing/voice regression.** `handleMessage` envelope validation, `seal`/`open` seq replay protection, machineId routing check, and `send()` are byte-identical; no voice paths touched.
- **No id/secret leaks.** New error strings are generic (`connection lost`, `too many pending requests`, `client closed`); the machineId-bearing timeout message is pre-existing and machineId is user-visible config, not a `pp_*`/session/pane id.
- **CI:** suite green on this HEAD (run 33176602350, 5m15s, all 32 checks PASS).

Fix B1 (single post-await guard + guard the catch-path `setState`) and this is approvable as-is.

umeranjum17:
## Re-review — verdict: **APPROVE**

Reviewed HEAD: `43ddaae70f14db7be562ac7477f8880454e4769f` against base `feat/compact-agent-activity` (merge-base `d47529ee`). Follow-up to [the prior REQUEST CHANGES review](https://github.com/umeranjum17/muxr/pull/155#issuecomment-5453409469). No blocking findings.

### Prior blocker B1 — resolved

`muxrClient.ts` now fences both resumption points of the awaited grant/ticket section:

- **L156** `if (this.closed || this.socket !== undefined) return;` before `new WebSocket(url)` — close-during-await can no longer create a socket, send `client.hello`, or resurrect `open` state; a concurrent open can no longer overwrite a live socket (prior N1 orphan closed by the same check).
- **L135** same guard at the top of the `catch` — a post-close/superseded failure can no longer `setState('stale'|'closed')`, fire `onTicketRejected`/`onPermanentError`, or schedule a reconnect.
- No stuck state from the guard: the only way `this.socket` is non-undefined at L156 is a socket whose `onclose` has not yet run; when it does, `retireSocket` reclaims ownership and schedules the reconnect. Prior N2 (stale doc-comment, L29) and N3 (guard parens, L104–105) also addressed.

**New delayed-ticket/grant assertions verified** (`connectionLoad.integration.test.ts` L172–238): the second test holds `issueWsTicket`'s fetch and `refreshHostedGrant` open across `close()`, then asserts zero WebSocket constructions, state pinned at `closed` with **no further state emissions**, and `onPermanentError` never called even when the parked grant rejects with "grant expired". Both assertions fail against the previous HEAD (socket creation on ticket resolve; `stale`+`onPermanentError` on grant rejection), so this is genuine regression coverage for exactly the B1 races.

### New route-change (Tailscale/half-open) recovery — verified

- **Half-open retirement** (muxrClient L240–248): a request timeout with no newer authenticated host traffic on the same socket now emits `stale`, then `retireSocket` — pending rejected with timers cleared, `closed`, one backoff-scheduled reconnect — then closes the dead socket; the socket's own later `onclose` hits the identity fence, so no double scheduling. No reconnect loop: backoff doubles to 30 s cap and only a real `onopen` resets attempts. The `hostFrameRevision` fence still confines retirement to a genuinely silent route (one slow request among live traffic only rejects itself).
- **Foreground rebuild / fresh ticket** (`_layout.tsx` L317–327): native-only, credentialed, fires only on a non-active→active transition; `syncReconnect()` → `sync.reconnect()` closes the old client (safe now per B1 — its in-flight open aborts instead of resurrecting) and the next `ensureClient()` dials with a freshly issued ticket. Subscription cleaned up on effect teardown.
- **No lifecycle rewrite**: `catalog/`, `packages/`, `apps/host`, `apps/relay` all have an empty diff vs base; rebuild lands in `refreshCatalog`, which folds host truth (herd-tree `agentStatus` + `lifecycle.catalog`), and `mergeCatalogAgent` preserves cached `agentStatus`/`lifecycleStateSince` when a DTO carries none.
- **Terminal reattach/full repaint** (`TerminalScreen.tsx` L246): foreground now calls `repaint()` instead of `reconnect()`. This is the correct fix: `reconnectNow` early-returns when the socket reads `CONNECTING|OPEN` (OpenTerminal L172), which is precisely the half-open state after a route change — the old handler left a frozen terminal. `repaint()` unconditionally replaces the socket (deliberate close, `onclose` nulled so no retry race) and re-attaches, and herdr sends the complete screen on attach. `OpenTerminal.ts` L208 adds the same post-await guard as B1 for the terminal channel, covering both orderings of racing `connectSocket` calls.
- **Route-loss test segment** (test L130–163): kills the host while the client's relay socket stays OPEN — a faithful local stand-in for a dead network route — then asserts the exact `stale → closed → connecting` ordering, recovery against a restarted host, and a healthy 5-session catalog read, still followed by the 30.5 s wait across the relay's first server ping.

### Unchanged / no regression

- **E2EE/replay/routing**: `handleMessage` validation, `seal`/`open` seq replay protection, machineId routing, and `send()` are byte-identical; terminal hosted-envelope validation untouched.
- **Native voice**: no voice paths in the diff (5 files: `_layout.tsx`, `muxrClient.ts`, `OpenTerminal.ts`, `TerminalScreen.tsx`, the test).
- **No id/secret leaks**: no new error strings beyond the prior review's inventory; test fixtures use synthetic ids.
- **CI**: green on this HEAD (run 33179198322, 4m46s, no FAIL rows); "unit: all vitest flows" 14.7 s (base) → 42.8 s, confirming both new flows executed, with ample headroom under the 150 s row timeout.

### Non-blocking notes

- Foreground resume now tears down and rebuilds even a perfectly healthy connection (ticket + full catalog batch per resume). Deliberate and cheap enough, but if resume-frequency ever matters, gating on `client.state !== 'open'` plus a probe request would be the refinement point.
- `open()`'s top guard admits a CLOSING husk and relies on L156 + the husk's `onclose → retireSocket` to resume the reconnect chain — sound, but subtle enough to deserve a one-line comment.

Approved as-is at `43ddaae7`.

PR #159, opened 2026-08-28
fix(mobile): clear final Herd emulator blockers

umeranjum17:
## Problem
- real Herd session output emitted high-frequency session snapshots, repeatedly rerendering VirtualizedList surfaces until the dev client hit Maximum update depth
- that churn continuously restarted the one-second activity visibility timer, so a fully visible completed card never acknowledged its row

## Fix
- suppress output-only session metadata churn and stop refetching topology for metadata updates
- preserve stable notification/card snapshots while applying real lifecycle changes
- keep equivalent catalog/tree snapshots referentially stable
- extend the existing activity/session flow coverage

## Verification
- root + mobile typechecks
- focused activity/tree/session flows: 12 passed
- full `yarn run check`: 34/34 passed
- Android API 36 x86_64 debug/dev-client build and install
- actual emulator: offscreen activity remained, partially visible card remained, fully visible foreground card cleared after one second and stayed cleared across restart
- actual dev client stayed connected with no Maximum update depth recurrence

Stacked on `feat/compact-agent-activity`. Do not merge yet.

### DIFF
--- a/apps/mobile/sources/catalog/application/sessionSync.integration.spec.ts
+++ b/apps/mobile/sources/catalog/application/sessionSync.integration.spec.ts
@@ -71,6 +71,7 @@ vi.mock('@/utils/sessionUtils', () => ({
     getSessionAvatarId: (session: Session) => session.id,
 }));
 import { applyStatusToSession, sessionInfoToSession } from '../infrastructure/sessionMapping';
+import { applyHostInfoToAgent } from '../domain/agent';
 import { storage } from './storage';
 
 async function spawn(options: { modelMode?: string; effortLevel?: string }) {
@@ -192,6 +193,26 @@ describe('session sync flow', () => {
         expect(mapped.metadata?.paneId).toBe('%1');
         expect(mapped.metadata?.summary?.text).toBe('Stabilizing realtime voice');
         expect(mapped.metadata?.agentName).toBe('Maria');
+        const metadata = mapped.metadata;
+        if (metadata === null) throw new Error('mapped herdr session must have metadata');
+        const stableExisting: Session = {
+            ...mapped,
+            metadata: { ...metadata, terminalTitle: 'stable terminal title' },
+        };
+        const outputOnlyUpdate = sessionInfoToSession({
+            id: 'session-a',
+            paneId: '%1',
+            cwd: '/work/alpha',
+            path: '/work/alpha',
+            created: '2026-01-01T00:00:00Z',
+            modified: '2026-01-01T00:00:01Z',
+            messageCount: 0,
+            firstMessage: '',
+            displayName: 'Maria',
+            taskTitle: 'Stabilizing realtime voice',
+            terminalTitle: 'volatile terminal output',
+        });
+        expect(applyHostInfoToAgent(stableExisting, outputOnlyUpdate)).toBe(stableExisting);
 
         const lifecycle = (agentStatus: 'working' | 'done') => ({
             sessionId: mapped.id,
--- a/apps/mobile/sources/catalog/application/storage.ts
+++ b/apps/mobile/sources/catalog/application/storage.ts
@@ -1,6 +1,7 @@
 import { create } from 'zustand';
 import { useShallow } from 'zustand/react/shallow';
 import React from 'react';
+import deepEqual from 'fast-deep-equal';
 import type { Session, Machine, SessionAgentModesPatch } from '../infrastructure/storageTypes';
 import type { Settings } from './settings';
 import { settingsDefaults } from './settings';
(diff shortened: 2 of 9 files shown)


umeranjum17 (commented):
## Verdict: APPROVE

Reviewed at head `f32a4cb0891d374d7d18322305e3c5b7d05098bf` against `feat/compact-agent-activity` (merge-base `6bb7fb7a`). Nine-file diff, all callsites, and host-side event semantics reviewed. CI `suite` green on this exact SHA (run 33219396400, `head_sha` matches). Note: GitHub rejects a formal APPROVE review event from the PR author's own account; this SHA-pinned review carries the explicit verdict instead.

## Blocker 1 — output-churn rerender storm / Maximum update depth: fixed at root

The loop was: host emits `session.updated` on stripped-terminal-title churn (`herdrSessionSource.emitState` info signature) → new `Session` object each event (`Date.now()` fallback + fresh `summary.updatedAt`) → `sessions` map identity churn → `useHerdTreeLive` subscriber refetches tree → new `herdrWorkspaces` array → pane/card/notification rebuilds → VirtualizedList commit storm. Every link is now cut at its source, not papered over:

- `parseTime(modified, createdAt)` stops self-inflicted `Date.now()` uniqueness per mapping.
- `applyHostInfoToAgent` returns the *existing reference* when the update differs only in `updatedAt`/`activeAt`/`presence` and volatile chrome (`terminalTitle`/`taskTitle`/`summary`). Verified against host: `observedTaskTitle` derives taskTitle from live chrome for adopted sessions and pins it for `ours`, and mobile has zero presentation consumers of `metadata.terminalTitle` — the tree (`pane.taskTitle`, 5s focused poll + 1s debounced refresh on real session changes) remains the display truth, so no user-visible staleness. Full catalog re-sync on reconnect/refresh restores byte-exact host truth.
- `applySessions`/`applyHerdrTree` deepEqual guards return prior state on equivalence (first-load carve-out via `herdrTreeLoaded` is correct).
- `sync.ts` no longer refetches the tree on `session.updated`; topology still propagates via `session.created`/`removed` refetch, the `useHerdTreeLive` sessions-subscription (fires on any *meaningful* session change, including tabId/workspaceId moves carried in metadata), and the 5s focused poll.
- `LiveTerminalsRow` derives cards in render (removes the setState-in-effect double-commit); `reconcileLiveTerminalCards` is idempotent, so the render-phase ref write is safe under StrictMode double-invoke and discarded renders.
- `KernelNotifications` `setPresentation` is equality-gated with a functional update — structurally cannot loop even if `herd` identity churns. `sameNotification` covers all five `HerdNotificationState` fields; `names` is a joined string, so `===` is value comparison.

Meaningful updates all still propagate: `active` flips, `agentName`, `provider`, `paneId`, worktree, lifecycle (`status.update`/`lifecycle.update` paths untouched), connection status (`socketStatus` untouched), and any non-frozen metadata diff breaks `deepEqual` and carries the fresh timestamps/presence with it. Reconnect truth is not rewritten: `refreshCatalog` `replace=true` rebuilds wholesale from host, `completionTransition`'s offline-baseline preservation is unchanged, and offline/connecting notification modes pass the equality gate by `mode`/`eventKey`.

## Blocker 2 — activity acknowledgement timer: fixed at root

Timer effect deps are now all stable under output churn: `cards` (reference-stable via reconcile, exhaustive 7-field compare of `LiveTerminalOrderCard`), `activityRows` (memo on `lifecycleEvents`/`ready`/`seenEventIds`, none of which move on output), `markSeen` (stable `useCallback([])`). Remaining restart triggers — scroll, layout, focus, foreground, real card changes — are exactly the relevant ones. Ack gating is correct: `visibilityTop`/`screenFocused`/`foreground` guards before arming, `visibleActivityEventIds` requires full vertical strip containment *and* full horizontal card containment (math matches `STRIP_GUTTER`/`CARD_WIDTH`/`CARD_GAP`/`snapToInterval`; fixed card dimensions; slot order preserved by reconcile), and the `measureInWindow` callback re-checks cancellation + `AppState` — no stale-closure ack race (React runs cleanup before re-effect). Event identity stays `eventId`-keyed; persistence unchanged: bounded 128-entry set, serialized AsyncStorage write chain — matches the emulator evidence of clears surviving restart.

## Fabric SIGSEGV assessment

No native code touched; zero new try/catch or error suppression anywhere in the diff — all churn suppression is strict-equality-gated, so it cannot hide divergent data, only identical data. The change *reduces* Fabric commit pressure by removing the update storm. A transient dev-client Fabric SIGSEGV on an emulator is consistent with known RN dev-client instability and is not attributable to, nor maskable by, this diff.

## Tests

New assertions extend existing flow specs (repo policy respected): the sessionSync spec drives the real `sessionInfoToSession → applyHostInfoToAgent` path with a volatile-terminal-title + timestamp-only update and asserts reference identity; the liveTerminalOrder spec asserts reconcile identity on equivalent snapshots. Both deterministic. `fast-deep-equal@^3.1.3` is properly declared in `apps/mobile/package.json`.

## Non-blocking notes

1. `applyHostInfoToAgent` spreads `...fresh`, which carries `draft: null` and clobbers a typed draft on the next meaningful update. Pre-existing in base (same spread); the equality gate actually narrows the exposure. Consider preserving `draft` alongside `agentState` in a follow-up.
2. The field-list equality helpers (`sameNotification`, reconcile's 7-field compare) are exhaustive today but will silently swallow any field added to their interfaces later without a compile error. Adjacent declarations mitigate; an exhaustiveness-checked compare would future-proof.
3. Removing the explicit `latestUsage` preservation is safe only because `sessionInfoToSession` never emits the key — worth a one-line comment so a future `latestUsage` in the DTO doesn't silently wipe usage.

PR #161, opened 2026-08-29
Fix mobile Stop agent routing

umeranjum17:
## What

Every explicit mobile layout action now maps to exactly its named scope:

- **Stop agent**: resolve one recognized live pane from the stable Agent Route after a fresh Herdr snapshot. Close that pane when its tab has siblings; otherwise close that tab only when the workspace has other tabs.
- **Close pane**: issue exactly one `pane.close`; refuse without mutation if the tab could not remain.
- **Close tab**: issue exactly one `tab.close`; refuse without mutation if the workspace could not remain.
- **Close workspace**: issue exactly one `workspace.close`; refuse without mutation when Herdr would widen a parent workspace close to its worktree group.
- **Close worktree group**: available only through Herdr's explicit group action and confirmation; never inferred from another action.

Targets never derive from cwd, name, title, or kind. No action closes multiple panes/tabs, a workspace, or a worktree group implicitly. Route-owned plugin streams, status, lifecycle, attachments, and identity clean up only after the exact selected close succeeds. Dead terminal `detachSession` surfaces are removed and fake mode closes only the selected session row.

## Why

Physical build45 widened mobile Stop into `Close worktree group?`. The contract is now strict: the UI label determines the maximum mutation scope, and structurally wider closes fail safely.

## Verify

```bash
yarn vitest run apps/host/src/requests/application/createRequestDispatcher.test.ts
yarn typecheck
yarn --cwd apps/mobile typecheck
yarn check:architecture
yarn check
node apps/host/dist/agent/infrastructure/layoutSelfCheck.js
```

- Dispatcher flows: 9/9 passed, including exact pane/tab/workspace forwarding and fake selected-row preservation.
- Infrastructure self-check records exact Stop/pane/tab/workspace Herdr RPCs, rejects hidden pane→tab, tab→workspace, and workspace→group widening with zero mutation, and verifies cleanup only after success.
- Root/mobile typechecks, architecture checks, and full repository suite passed.

Do not merge or publish from this PR.

### DIFF
--- a/apps/host/src/agent/application/openTerminal.ts
+++ b/apps/host/src/agent/application/openTerminal.ts
@@ -15,7 +15,6 @@ export type OpenTerminalResult =
 export interface TerminalPort {
     attach(command: OpenTerminalCommand): Promise<{ paneId: string }>;
     detach(channel: string, deviceId?: string): void;
-    detachSession?(sessionId: string): void;
 }
 
 export async function openTerminal(port: TerminalPort | undefined, command: OpenTerminalCommand): Promise<OpenTerminalResult> {
--- a/apps/host/src/agent/application/stopAgent.ts
+++ b/apps/host/src/agent/application/stopAgent.ts
@@ -9,12 +9,10 @@ export type StopAgentResult = { ok: true; data: null } | { ok: false; error: str
 
 export interface StopAgentPorts {
     sessions: Pick<SessionSource, 'stop' | 'abort' | 'reload'>;
-    detachSession?: (sessionId: string) => void;
 }
 
 export async function stopAgent(ports: StopAgentPorts, command: StopAgentCommand): Promise<StopAgentResult> {
     if (command.action === 'stop') {
-        ports.detachSession?.(command.sessionId);
         await ports.sessions.stop(command.sessionId);
         return { ok: true, data: null };
     }
--- a/apps/host/src/agent/infrastructure/fakeSessionSource.ts
+++ b/apps/host/src/agent/infrastructure/fakeSessionSource.ts
@@ -334,7 +334,9 @@ export function createFakeSessionSource(): SessionSource {
 
         async abort() {},
         async stop(sessionId) {
+            requireSession(sessionId);
             sessions.delete(sessionId);
+            emit(sessionId, { type: 'session.removed' });
         },
         async reload() {},
         subscribe(listener) {
--- a/apps/host/src/agent/infrastructure/herdrSessionSource.ts
+++ b/apps/host/src/agent/infrastructure/herdrSessionSource.ts
@@ -171,6 +171,113 @@ export async function sendKeysToLiveAgent(
     await client.call('agent.send_keys', { target: record.paneId, keys });
 }
 
+function agentRouteError(code: 'agent-unavailable' | 'agent-route-ambiguous'): Error {
+    const message = code === 'agent-route-ambiguous'
+        ? 'That Agent Route is ambiguous. Refresh and select the agent again.'
+        : 'That agent is no longer available. Refresh and try again.';
+    return Object.assign(new Error(message), { code });
+}
+
+/** Close one pane only when Herdr can preserve its tab and workspace. */
+export async function closeExactPane(
+    client: Pick<HerdrClient, 'call'>,
+    paneId: string,
+    panes: ReadonlyMap<string, Pick<PaneRecord, 'tab_id'>>,
+): Promise<void> {
+    const tabId = panes.get(paneId)?.tab_id;
+    if (tabId === undefined) {
+        throw Object.assign(new Error('That pane is no longer available. Refresh and try again.'), { code: 'pane-unavailable' });
+    }
+    let paneCount = 0;
+    for (const pane of panes.values()) {
+        if (pane.tab_id === tabId) paneCount += 1;
+    }
+    if (paneCount <= 1) {
+        throw Object.assign(new Error('Closing this pane would also close its tab. Use Close tab instead.'), { code: 'pane-close-would-widen' });
+    }
+    await client.call('pane.close', { pane_id: paneId });
+}
+
+/** Close one tab only when Herdr can preserve its workspace. */
+export async function closeExactTab(
+    client: Pick<HerdrClient, 'call'>,
+    tabId: string,
+    tabs: ReadonlyMap<string, Pick<TabRecord, 'workspace_id'>>,
+): Promise<void> {
+    const workspaceId = tabs.get(tabId)?.workspace_id;
+    if (workspaceId === undefined) {
+        throw Object.assign(new Error('That tab is no longer available. Refresh and try again.'), { code: 'tab-unavailable' });
+    }
+    let tabCount = 0;
+    for (const tab of tabs.values()) {
+        if (tab.workspace_id === workspaceId) tabCount += 1;
+    }
+    if (tabCount <= 1) {
+        throw Object.assign(new Error('Closing this tab would also close its workspace. Use Close workspace instead.'), { code: 'tab-close-would-widen' });
+    }
+    await client.call('tab.close', { tab_id: tabId });
+}
+
+/** Close one workspace, refusing Herdr's implicit parent-worktree group widening. */
+export async function closeExactWorkspace(
+    client: Pick<HerdrClient, 'call'>,
+    workspaceId: string,
+    workspaces: ReadonlyMap<string, WorkspaceRecord>,
+): Promise<void> {
+    const workspace = workspaces.get(workspaceId);
+    if (workspace === undefined) {
+        throw Object.assign(new Error('That workspace is no longer available. Refresh and try again.'), { code: 'workspace-unavailable' });
+    }
+    const worktree = workspace.worktree;
+    if (worktree?.is_linked_worktree === false && worktree.repo_key !== undefined) {
+        let groupSize = 0;
+        for (const candidate of workspaces.values()) {
+            if (candidate.worktree?.repo_key === worktree.repo_key) groupSize += 1;
+        }
+        if (groupSize >= 2) {
+            throw Object.assign(new Error('Closing this workspace would close its worktree group. Use the explicit Close worktree group action instead.'), {
+                code: 'worktree-group-confirmation-required',
+            });
+        }
+    }
+    await client.call('workspace.close', { workspace_id: workspaceId });
+}
+
+/** Close exactly the live pane or sole-pane tab selected by one stable Agent Route. */
+export async function closeAgentRoute(
+    client: Pick<HerdrClient, 'call'>,
+    sessionId: string,
+    identities: readonly Pick<AgentIdentity, 'sessionId' | 'paneId'>[],
+    liveAgents: Pick<ReadonlyMap<string, unknown>, 'has'>,
+    panes: ReadonlyMap<string, Pick<PaneRecord, 'tab_id'>>,
+    tabs: ReadonlyMap<string, Pick<TabRecord, 'workspace_id'>>,
+    afterClose: (record: Pick<AgentIdentity, 'sessionId' | 'paneId'>) => void,
+): Promise<void> {
+    let match: Pick<AgentIdentity, 'sessionId' | 'paneId'> | undefined;
+    for (const identity of identities) {
+        if (identity.sessionId !== sessionId) continue;
+        if (match !== undefined) throw agentRouteError('agent-route-ambiguous');
+        match = identity;
+    }
+    if (match === undefined || !liveAgents.has(match.paneId)) throw agentRouteError('agent-unavailable');
+    const tabId = panes.get(match.paneId)?.tab_id;
+    if (tabId === undefined) throw agentRouteError('agent-unavailable');
+    let paneCount = 0;
+    for (const pane of panes.values()) {
+        if (pane.tab_id === tabId) paneCount += 1;
+    }
+    if (paneCount === 0) throw agentRouteError('agent-unavailable');
+    try {
+        if (paneCount > 1) await closeExactPane(client, match.paneId, panes);
+        else await closeExactTab(client, tabId, tabs);
+    } catch (error) {
+        if (error !== null && typeof error === 'object' && 'code' in error
+            && (error.code === 'pane-close-would-widen' || error.code === 'tab-close-would-widen')) throw error;
+        throw agentRouteError('agent-unavailable');
+    }
+    afterClose(match);
+}
+
 
 interface PaneRecord {
     pane_id: string;
@@ -191,7 +298,13 @@ interface WorkspaceRecord {
     workspace_id: string;
     label?: string;
     focused?: boolean;
-    worktree?: { repo_name?: string; repo_root?: string; checkout_path?: string };
+    worktree?: {
+        repo_key?: string;
+        repo_name?: string;
+        repo_root?: string;
+        checkout_path?: string;
+        is_linked_worktree?: boolean;
+    };
 }
 
 interface TabRecord {
@@ -838,9 +951,7 @@ export async function createHerdrSessionSource(
     }
 
     function agentUnavailable(): Error {
-        const error = new Error('That agent is no longer available. Refresh and try again.') as Error & { code: string };
-        error.code = 'agent-unavailable';
-        return error;
+        return agentRouteError('agent-unavailable');
     }
 
     async function resolvePane(sessionId: string): Promise<AgentIdentity> {
@@ -1875,16 +1986,19 @@ export async function createHerdrSessionSource(
 
         async closeTab(sessionId: string, tabId: string): Promise<void> {
             await resolvePane(sessionId);
-            await client.call('tab.close', { tab_id: tabId });
+            await closeExactTab(client, tabId, tabsById);
         },
 
         async closePane(sessionId: string): Promise<void> {
             const record = await resolvePane(sessionId);
-            await client.call('pane.close', { pane_id: record.paneId });
+            await closeExactPane(client, record.paneId, panesById);
         },
 
         async closeWorkspace(workspaceId: string): Promise<void> {
-            await client.call('workspace.close', { workspace_id: workspaceId });
+            try { await refreshSnapshot(); } catch {
+                throw Object.assign(new Error('That workspace is no longer available. Refresh and try again.'), { code: 'workspace-unavailable' });
+            }
+            await closeExactWorkspace(client, workspaceId, workspacesById);
         },
 
         async createTab(sessionId: string, options: { kind?: string; label?: string }): Promise<void> {
@@ -1943,17 +2057,20 @@ export async function createHerdrSessionSource(
         },
 
         async stop(sessionId: string): Promise<void> {
-            pluginStreams?.closeSession(sessionId);
-            const record = await resolvePane(sessionId);
-            await client.call('pane.close', { pane_id: record.paneId }).catch(() => {});
-            // Close the pane's status watch BEFORE removing the identity record --
-            // the syncDiscovery cleanup only walks live records, so a watch
-            // orphaned here leaks a socket (and its reconnect loop) forever.
-            statusWatches.get(record.paneId)?.();
-            statusWatches.delete(record.paneId);
-            identity.remove(sessionId);
-            options.lifecycle?.remove(sessionId);
-            clearAttention(sessionId);
+            try { await refreshSnapshot(); } catch { throw agentUnavailable(); }
+            await closeAgentRoute(client, sessionId, identity.all(), agentsByPane, panesById, tabsById, (record) => {
+                pluginStreams?.closeSession(record.sessionId);
+                statusWatches.get(record.paneId)?.();
+                statusWatches.delete(record.paneId);
+                lifecycleEpochByPane.delete(record.paneId);
+                lastStateSignature.delete(record.sessionId);
+                lastInfoSignature.delete(record.sessionId);
+                attachments.dropPane(record.paneId);
+                identity.remove(record.sessionId);
+                options.lifecycle?.remove(record.sessionId);
+                clearAttention(record.sessionId);
+                publish(record.sessionId, { type: 'session.removed' });
+            });
         },
 
         async abort(sessionId: string): Promise<void> {
--- a/apps/host/src/agent/infrastructure/layoutSelfCheck.ts
+++ b/apps/host/src/agent/infrastructure/layoutSelfCheck.ts
@@ -13,7 +13,7 @@ import {
 import { lifecycleReasonForObservation } from '../domain/lifecycle.js';
 import { IdentityStore, parseTaskTitle } from './identity.js';
 import { RealtimeCodingCoordinator } from './realtimeCoordinator.js';
-import { sendKeysToLiveAgent } from './herdrSessionSource.js';
+import { closeAgentRoute, closeExactPane, closeExactTab, closeExactWorkspace, sendKeysToLiveAgent } from './herdrSessionSource.js';
 import { createConnection } from 'node:net';
 import { mkdtempSync, writeFileSync } from 'node:fs';
 import { tmpdir } from 'node:os';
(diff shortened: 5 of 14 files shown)


umeranjum17:
## REQUEST CHANGES — PR #161 @ `3fbae911690eab1d0dee197e7f7753830a374809`

Read-only strict review of the amended HEAD. Full diff read (9 files, +118/-39) against `main` = `8cf722275ad75e524cfcbc44b041357996c80281`; merge-base equals main's tip, so no rebase drift. CI at this SHA is green: `suite` ✅ (https://github.com/umeranjum17/muxr/actions/runs/33234117333/job/99051964361), CodeQL ✅. PR: https://github.com/umeranjum17/muxr/pull/161

### Verified correct — the core routing fix holds

- `herdrSessionSource.stop()` (L1971-1975): fresh `refreshSnapshot()` (which reconciles identity via `syncDiscovery()`, L733) → `interruptAgentRoute(client, sessionId, identity.all(), agentsByPane)`. A failed snapshot throws `agent-unavailable` before any RPC.
- `interruptAgentRoute` (L182-200) matches on `sessionId` only. No `cwd`, `foreground_cwd`, `terminal_title`, agent name, label, or kind anywhere in the path. It requires the pane to be present in `agentsByPane` — a recognized live **agent** pane, strictly stronger than the old `resolvePane` `panesById` check — then issues exactly one `agent.send_keys { target: paneId, keys: ['escape'] }`.
- No `pane.close`, `tab.close`, `workspace.close`, `worktree.*`, ctrl+c, SIGINT, or kill anywhere in the stop path. The only `pane.close` sites left in the file are squad rollback (L908), start-failure cleanup (L1020), and the explicit `closePane` port (L1909).
- Stop no longer performs stale-identity cleanup, `lifecycle.remove`, `clearAttention`, or `statusWatches` teardown. Keeping the pane's status watch is right, since the pane now survives.
- Terminal detachment is gone at the callsite: `createRequestDispatcher.ts` L189-191 passes `{ sessions: source }` only, and `StopAgentPorts` lost `detachSession`. The selected terminal stream stays attached.
- No false success. `useCaseData` rethrows with `code`, `fromCaught` maps it onto the frame, `normalizeRequestFailure` passes the message through unchanged, `MuxrRequestError` surfaces in `Modal.alert('Stop failed', …)`. Copy leaks no pane id or route (asserted in the new dispatcher test).
- Stop and explicit Close stay separate. Mobile Stop: `TerminalScreen.stopSession` → `sessionStop` (ops) → mobile `stopAgent` use case → `session.stop`. The only `pane.close` / `workspace.close` callers are `SpacesTree.tsx` L404 / L375 behind "Close pane?" / "Close space?". `sessionKill` is fully renamed with zero remaining references.
- Infrastructure self-check runs in CI (`scripts/diagnostics/application/runSuite.mjs` L27 → `layoutSelfCheck.js`) and pins one escape call plus zero mutations on stale and ambiguous routes. The dispatcher test is one flow-level test through the real dispatcher with a fake port (7/7 in the file), consistent with the repo's test policy.

### Blockers

1. **`docs/ARCHITECTURE.md:128` still documents the removed behavior**: `| session.stop | close the pane | pane.close |`. The canonical contract table now contradicts the new `RequestMap['session.stop']` doc comment. This is exactly the mapping a future contributor would follow to reintroduce the build45 regression.
2. **Dead removed-feature surface left in tree.** `TerminalManager.detachSession` (`apps/host/src/agent/infrastructure/terminalManager.ts:352-357`) has zero callers at this SHA, and its docstring still claims "session.stop, host shutdown" — shutdown uses `closeAll()` (L359). `TerminalPort.detachSession?` (`apps/host/src/agent/application/openTerminal.ts:18`) is likewise unreferenced. Clean cutover means deleting both; leaving them invites Stop being wired back into stream teardown.
3. **`fakeSessionSource.stop` still deletes the session** (`apps/host/src/agent/infrastructure/fakeSessionSource.ts:336-338`). The `--fake` host is the harness used to exercise mobile without herdr, and it keeps modelling Stop as "the row disappears" — the precise semantics this PR forbids. Make the fake keep the session so fake-port behavior matches the contract this commit defines.
4. **One non-escape mutation survives in `stop()`**: `pluginStreams?.closeSession(sessionId)` (L1974) closes every plugin stream attachment bound to the route — child process plus WebSocket, reason `'session stopped'` (`pluginStreamManager.ts:432-436`). Now that the pane, siblings, and terminal all survive, a Stop tap still tears down that agent's realtime/voice provider stream. Either drop it or state it explicitly in the `session.stop` contract comment; as written the PR claims Stop sends only escape while still tearing down provider streams. (Failure paths are clean — this runs only after a successful send.)

### Non-blocking findings

5. `interruptAgentRoute` scans `identity.all()` (a fresh array copy of every identity, on every Stop) and guards ambiguity, but `IdentityStore` is a `Map` keyed by `sessionId`, so duplicate routes are structurally impossible. `agent-route-ambiguous` is therefore unreachable in production, and both the self-check and the dispatcher test reach it only via hand-built duplicate arrays. `identity.get(sessionId)` is exact, O(1), and allocation-free.
6. The `catch` at L197-199 maps every RPC/transport failure to `agent-unavailable`, so a reachable-but-erroring herdr tells the user "That agent is no longer available. Refresh and try again." and the diagnostics journal records the wrong code.
7. Stop is offered unconditionally in the terminal action sheet (`TerminalScreen.tsx:758`). Shell panes and panes whose agent already exited both keep a session but are absent from `agentsByPane`, so Stop on them now fails with "That agent is no longer available" instead of a truthful message. Consider gating the row on a live agent.
8. Coverage shape: the self-check exercises the helper directly, the dispatcher test exercises the dispatcher with a fake port. The composed `herdrSessionSource.stop()` — refresh → interrupt → `closeSession`, and no layout RPC — is the actual regression site and has no automated guard. One flow-level assertion with a fake herdr client would close that gap without growing a suite.
9. API 36 evidence is self-reported prose only: no logs, screenshots, or build artifacts are attached to the PR, so the physical-emulator claims (isolated API 36 install, 8 tabs / 8 panes intact, no close-group warning) are not verifiable from the repository. The automated half is covered by the green `suite` run at this SHA.

Nothing merged, published, or edited. Blockers 1-4 are cheap and mechanical; the routing core itself needs no rework.

umeranjum17:
## Supersedes my previous review — Stop semantics corrected by owner

My earlier comment on `3fbae911690eab1d0dee197e7f7753830a374809` validated the escape-only Stop design ("core routing fix holds"). **That endorsement is withdrawn.** The owner has corrected the required semantics:

- **Stop agent = close only the selected pane**, or **close the selected tab when the route's pane is that tab's sole pane**.
- **Escape stays a separate Send Keys action** — never the meaning of Stop.

Verdict on this SHA is unchanged in outcome and stronger in reason: **REQUEST CHANGES**. The escape-only `session.stop` in `3fbae911` (`herdrSessionSource.ts` L1971-1975, `interruptAgentRoute` L182-200, contract comment "never closes layout") implements the wrong use case and must not be approved or merged as-is. HEAD is still `3fbae911` at the time of writing; I am holding for the owner-amended HEAD and will re-review only that SHA.

### Destructive boundary — established from herdr 0.8.0 itself, not inferred

`herdr api schema --json` (protocol 20) and the shipped binary give the authoritative contract:

- `pane.close` takes `PaneTarget { pane_id }` only. `tab.close` takes `TabTarget { tab_id }` only. **Neither has a force or confirm parameter.**
- herdr refuses the destructive escalation itself, with error code `confirmation_required` and messages `closing this pane would close a worktree group` / `closing this tab would close a worktree group`. The `Close worktree group?` / `Close group` / `Delete worktree checkout...` / `delete anyway` strings are the desktop client's confirm path.
- The only socket call that can actually destroy a checkout is `worktree.remove { workspace_id, force }` — `force` defaults to `false`. `workspace.close { workspace_id }` closes a whole space.

So the amended Stop is safe **iff** it issues exactly one of `pane.close` / `tab.close` and treats `confirmation_required` as a terminal, non-retryable refusal.

### What I will verify on the amended HEAD

**Targeting (must be identical to the good part of this PR)**
1. Fresh `refreshSnapshot()` before resolving; snapshot failure → fail with no RPC.
2. Route resolved from the stable Agent Route alone — exactly one identity, pane present in `agentsByPane`. No cwd / `foreground_cwd` / `terminal_title` / agent name / label / kind inference anywhere.
3. Sole-pane determination computed **host-side from the post-refresh snapshot** (`panesById` filtered by the route pane's `tab_id`), never from the mobile `siblings` state in `TerminalScreen`, never from cached catalog tree data.

**Layout-close boundary**
4. Exactly one destructive RPC per Stop: `pane.close` when siblings remain, `tab.close` when the route's pane is the tab's only pane. Never both, and no escalation chain (`pane.close` failure must not fall back to `tab.close`).
5. Zero `workspace.close`, zero `worktree.remove`, zero `force: true`, anywhere in the Stop path.
6. `confirmation_required` from herdr is surfaced as a truthful failure with no retry and no auto-confirm. Better: pre-check the snapshot (`workspace.worktree.checkout_path` on the route's workspace plus a sole-tab test) and refuse before sending, so the desktop never gets a modal a phone tap caused.
7. No `agent.send_keys` and no `pane.send_keys` in the Stop path at all — Escape belongs to Send Keys/abort, not Stop.

**Teardown that must come back (it was correct before this PR and the diff deleted it)**
8. `pluginStreams.closeSession(sessionId)`, then close-and-delete `statusWatches.get(paneId)` **before** `identity.remove(sessionId)` — the deleted comment was load-bearing: `syncDiscovery` only walks live records, so a watch orphaned here leaks a socket and its reconnect loop forever.
9. `identity.remove`, `options.lifecycle?.remove`, `clearAttention`, and terminal detachment (`terminals.detachSession` — currently deleted from `StopAgentPorts` and now dead code in `TerminalManager`) all belong back in the stop path: the pane is gone, so an attached terminal stream to a dead pane is a leak, not a feature.
10. Mobile must navigate off the closed pane again (the removed `siblings` → `router.replace` / `router.back` logic). Sitting on a terminal whose pane no longer exists is a dead screen.

**Escape as a separate Send Keys action**
11. Today the only mobile path that delivers Escape is the voice plugin's `send_agent_keybinding` (allowlist `['escape']` → coordinator `key` → `sendKeys`). `sessionAbort` exists in `catalog/application/ops.ts` but has **zero UI callers** at this SHA, and the terminal key toolbar has no Esc cap. If Stop goes back to closing layout, the amended HEAD must expose the separate interrupt/Send-Keys affordance, or mobile silently loses "interrupt this turn without closing anything".

**Copy, docs, fakes, coverage**
12. `RequestMap['session.stop']` comment must be rewritten — "never closes layout" becomes false. `docs/ARCHITECTURE.md:128` must state pane close *and* the sole-pane→tab-close rule.
13. Mobile Stop modal copy must name exactly what closes (this pane; this tab when it is the last pane) and must drop "its herdr pane, sibling panes, and workspaces stay open".
14. `fakeSessionSource.stop` deleting the session becomes correct again under this design — keep it consistent with whichever contract ships.
15. Automated coverage must pin the two branches and the refusal: sibling case → exactly one `pane.close`; sole-pane case → exactly one `tab.close`; worktree-group case → zero close RPCs; stale/ambiguous route → zero RPCs; and no `workspace.close`/`worktree.remove` in any branch. The existing `layoutSelfCheck` fake-client pattern (it runs in CI via `runSuite.mjs` L27) plus the dispatcher fake-port test are the right two places — no new suite needed.

**Evidence**
16. Physical evidence for the amended HEAD should show the tab-close branch on a worktree-backed workspace explicitly: pane count before/after, tab count before/after, and that no `Close worktree group?` dialog appeared on the desktop. The current PR body's emulator evidence covers only the escape design and no longer applies.

No approval will be issued for `3fbae911`. Ping me when the amended HEAD is pushed and I will re-review against exactly the boundaries above.

umeranjum17:
## APPROVE — PR #161 @ `1d3c6199e4855156c47aaa697a571494435968b6`

Re-reviewed the amended HEAD in full (13 files, +192/-41 against `main` = `8cf722275ad75e524cfcbc44b041357996c80281`; merge-base equals main's tip). CI at this exact SHA: `suite` ✅ (https://github.com/umeranjum17/muxr/actions/runs/33235419962/job/99055439420), CodeQL ✅, both Analyze jobs ✅. Focused dispatcher file now holds 8 `it` blocks, matching the 8/8 claim. Nothing merged or published.

This supersedes both of my prior comments: the escape-only design is gone, and every blocker I raised is resolved.

### Exact-close boundary — verified against the checklist

**Targeting.** `stop()` (`herdrSessionSource.ts` L1982-1997) refreshes the snapshot first (failure → `agent-unavailable`, zero RPC), then delegates to `closeAgentRoute` (L182-211). The route resolves from the stable Agent Route alone — exactly one identity, pane must be present in `agentsByPane`. No cwd, `foreground_cwd`, `terminal_title`, agent name, label, or kind appears anywhere in the path.

**Branch selection is host-side and snapshot-derived.** The tab comes from `panesById.get(match.paneId).tab_id`, and the sibling count is computed by walking `panesById` for that `tab_id` (L197-202) — not from `TerminalScreen`'s `siblings` state, not from cached tree data. I checked the undercount risk this would have if `tab_id` were ever absent: herdr 0.8.0's `session.snapshot` schema makes `panes` required on `SessionSnapshot` and `tab_id` **required** on `PaneInfo` (`herdr api schema --json`, protocol 20), and the snapshot is global across workspaces, so the per-tab count is complete and the `tabId === undefined` guard is purely defensive. A mis-targeted `tab.close` on a tab that still holds a sibling is not reachable.

**Exactly one close RPC, correct branch.** `paneCount > 1` → one `pane.close { pane_id }`; otherwise one `tab.close { tab_id }` (L204-206). No both-branch, no escalation chain, no retry. Repo-wide inventory at this SHA confirms Stop touches nothing else: `workspace.close` exists only in the explicit `closeWorkspace` port (L1923-1925) behind the Herd screen's "Close space?", `worktree.create` only in `start`, and `worktree.remove` appears nowhere in the host. No `agent.send_keys` / `pane.send_keys` in the Stop path.

**Group boundary.** herdr itself is the backstop: `pane.close` takes only `PaneTarget { pane_id }`, `tab.close` only `TabTarget { tab_id }`, neither has a force or confirm parameter, and the binary refuses with code `confirmation_required` plus `closing this pane would close a worktree group` / `closing this tab would close a worktree group`. The only socket call that could destroy a checkout is `worktree.remove { workspace_id, force }`, which this PR never issues. muxr cannot drive the `Close worktree group?` path.

**Cleanup strictly after a successful exact close.** `afterClose(match)` runs only past the awaited RPC (L210), and the ordering inside the callback keeps the load-bearing invariant: `pluginStreams.closeSession` → close and delete `statusWatches[paneId]` → epoch/signature drops → `attachments.dropPane` → `identity.remove` → `lifecycle.remove` → `clearAttention` → `publish session.removed`. The status watch is torn down **before** the identity record, so no watch is orphaned outside `syncDiscovery`'s live-record walk. Every failure path (stale route, ambiguous route, refused or failed close) throws before `afterClose`, leaving streams, watches, identity, lifecycle, and attention untouched — and the self-check asserts exactly that, including a `failStopClose` case.

**Terminal detachment.** I had asked for `detachSession` to come back; removing it (both `TerminalManager.detachSession` and `TerminalPort.detachSession?`) is the better call and I verified the mechanism rather than taking it on trust: each attachment is a spawned herdr stream child, and `child.on('exit')` → `finish(...)` emits `terminal.closed`, closes the socket, and deletes the attachment (`terminalManager.ts` L320-322, L192-215). Pane death terminates the stream on its own, so there is no leak and no dead-pane stream held open. Zero `detachSession` references remain anywhere.

**Escape stayed separate.** `session.abort` is untouched (`agent.send_keys ['escape']` with a `pane.send_keys` fallback), and the voice policy still routes every interrupt/cancel/escape request through `send_agent_keybinding` with the `['escape']` allowlist. Stop no longer has any key-sending behavior to confuse with it.

**Contract, docs, UI, fake.** `RequestMap['session.stop']` now documents pane-or-sole-pane-tab close and the workspace/worktree prohibition. `docs/ARCHITECTURE.md:128` states the same rule with both RPCs and the explicit "never `workspace.close` or worktree/group close". Mobile copy names exactly what closes and what survives, and the pane-successor navigation is back with `siblings` in the dependency list (`TerminalScreen.tsx` L325-348). `fakeSessionSource.stop` now guards with `requireSession`, removes only the selected session, and emits `session.removed`; the new dispatcher test proves the sibling survives, the selected route's `open` rejects, and the event fires. `session.stop` remains outside `VIEW_ONLY_REQUESTS`, so observer/browser grants still cannot close layout.

### Non-blocking follow-ups

1. **`confirmation_required` is reported as `agent-unavailable`.** The `catch` at L207-208 collapses every close failure — including herdr's worktree-group refusal and plain transport errors — into "That agent is no longer available. Refresh and try again." The safety property holds (no group close, no cleanup, no retry), but the message is untrue in the refusal case. Passing herdr's code and message through, or pre-checking `workspace.worktree.checkout_path` plus a sole-tab test, would make it honest.
2. **Stop on a pane with no live agent now refuses.** `liveAgents.has` means a shell pane or a pane whose agent has exited fails with "no longer available", where `main` closed it. That matches the "selected live Agent Route" wording, but the Stop row in the terminal action sheet is still rendered unconditionally (`TerminalScreen.tsx` L764). Gating the row on a live agent, or wording that case separately, would avoid a confusing failure on a pane the user can plainly see.
3. **Dead guard.** `paneCount === 0` (L203) is unreachable: the matched pane is in `panes`, so the count is at least 1.
4. **Unreachable ambiguity branch (carried over).** `IdentityStore` is a `Map` keyed by `sessionId`, so `identity.all()` cannot contain duplicates; `agent-route-ambiguous` is reachable only via hand-built arrays in the checks, and `identity.get(sessionId)` would be exact, O(1), and allocation-free versus the per-Stop `identity.all()` copy. Cheap defensive code, but it is dead in production.
5. **Restore one line of rationale.** The deleted comment explaining *why* the status watch must close before `identity.remove` was load-bearing; the ordering is now correct but implicit. One sentence in the `afterClose` callback would protect it from a future reorder.
6. **Residual, by design.** Closing a sole-pane tab that is also its workspace's last tab leaves herdr to dispose the now-empty workspace; muxr issues no workspace RPC, and worktree-backed workspaces are refused by herdr instead. Worth stating in the physical-verification notes so the observed disappearance of an empty space is not read as an escalation.

Approved on substance for `1d3c6199e4855156c47aaa697a571494435968b6`. Items 1-6 are polish, not gates.

umeranjum17:
## APPROVE — PR #161 @ `39eb32de6f646f81c58a3ff8c2d8741b0ff69cfd`

Re-reviewed the final HEAD in full: 14 files, +383/-50 against `main` = `8cf722275ad75e524cfcbc44b041357996c80281` (merge-base equals main's tip), plus the incremental diff against the previously approved `1d3c6199` (7 files, +204/-22). CI at this exact SHA: `suite` ✅ (https://github.com/umeranjum17/muxr/actions/runs/33236143972/job/99057357683), CodeQL ✅, both Analyze jobs ✅. Dispatcher file now holds 9 `it` blocks, matching the 9/9 claim. Nothing merged or published.

The scope rule is enforced at each level, and the level boundaries are the right ones.

### Scope rule verified end to end

**Stop agent** (`closeAgentRoute`, L247-279). Fresh snapshot first; route resolved from the stable Agent Route only (exactly one identity, pane must be in `agentsByPane`); zero cwd/`foreground_cwd`/title/name/label/kind inference. Multi-pane tab → `closeExactPane`; sole-pane tab → `closeExactTab`, which refuses when that tab is its workspace's last tab. Widening codes (`pane-close-would-widen`, `tab-close-would-widen`) are re-thrown verbatim so the user sees the real reason; anything else collapses to `agent-unavailable`. `afterClose` still runs only past a successful RPC, and the callback keeps the load-bearing order — status watch closed and deleted before `identity.remove`.

**Close pane** (`closeExactPane`, L182-199) refuses at `paneCount <= 1`, i.e. pane→tab widening, before any RPC. Reached through `closePane`, which resolves the route via `resolvePane` (snapshot refresh included).

**Close tab** (`closeExactTab`, L202-219) refuses at `tabCount <= 1`, i.e. tab→workspace widening. Reached through `closeTab`, also refreshed via `resolvePane`.

**Close workspace** (`closeExactWorkspace`, L222-244) refreshes the snapshot itself, then refuses only for a **parent** checkout (`is_linked_worktree === false`) that still has at least one other workspace sharing its `repo_key` — precisely the parent→worktree-group widening case. Linked-worktree workspaces and ungrouped parents close normally, which the self-check pins with the `w6`/`w4` cases.

**Group close** exists nowhere in muxr. Repo-wide at this SHA, the only `pane.close` / `tab.close` / `workspace.close` client calls are inside these three helpers (plus two pre-existing internal cleanup calls, see note 2); `worktree.remove` appears nowhere in the host, and no code path passes `force`.

### Field-level evidence, checked against herdr 0.8.0 rather than assumed

- `WorkspaceWorktreeInfo` — the `worktree` field of snapshot `WorkspaceInfo` — **requires** `repo_key`, `repo_name`, `repo_root`, `checkout_path`, `is_linked_worktree`. The new `WorkspaceRecord` fields exist and are always populated when `worktree` is present, so the group guard cannot silently no-op.
- `PaneInfo` requires `tab_id`; `TabInfo` requires `workspace_id`. The per-tab pane count and per-workspace tab count are therefore complete, and the snapshot is global across workspaces, so neither count can be short.
- `WorkspaceInfo` requires `active_tab_id` and `tab_count`: a herdr workspace always holds at least one tab. That is exactly why refusing the last-tab close is the correct boundary rather than an over-restriction — the alternative genuinely is workspace disposal.
- Defense in depth is intact on the workspace path: `closeExactWorkspace` does not wrap its RPC, so if herdr itself answers `confirmation_required` ("closing this … would close a worktree group") the authoritative message and code reach the client unmasked.

### Existing consumers re-checked

- Mobile labels/handlers: "Close workspace?" and "Close pane?" now describe exactly what closes and state that nothing closes if it would widen; Stop's copy gained the same guarantee. Mobile still routes Stop through the shared use case and keeps the pane-successor navigation.
- `session.stop` remains outside `VIEW_ONLY_REQUESTS`, so observer/browser grants still cannot close layout at any level.
- The two live-herdr e2e checks (`checkHerdrE2E.mjs`, `checkWorktreeE2E.mjs`, both gated on a herdr socket in `runSuite.mjs` L44-45) end with `session.stop` on a freshly started agent. That still succeeds: `session.start` always `tab.create`s into a workspace herdr had already created with a root tab, so the agent's tab is never its workspace's only tab. [INFERENCE — derived from the herdr schema and the documented `workspace.create` / `worktree.create` responses; I did not mutate the live herdr to prove it.]

### Non-blocking follow-ups

1. **The phone has no Close tab action.** `closeExactPane`'s refusal says "Use Close tab instead", but `SpacesTree` exposes only workspace and pane closes — tab rows have no long-press close, and no mobile code calls `tab.close`. On mobile, a sole-pane tab can only be closed via Stop agent. Either add the tab-row affordance or word that message client-agnostically.
2. **Two internal cleanup paths still bypass the guard.** Squad rollback (L993) and start-failure cleanup (L1105) call `pane.close` directly on a pane that is the only pane of a tab muxr itself just created, so they implicitly close that tab. Self-owned scaffolding rather than user layout, but it is the one place where the codebase does what the rule forbids elsewhere; routing them through `closeExactTab` or noting why they are exempt would keep the invariant readable.
3. **`closeAgentRoute` still flattens non-widening failures** — `pane-unavailable`, `tab-unavailable`, transport errors, and herdr's own `confirmation_required` all become "That agent is no longer available." Safe (no mutation, no retry, no cleanup) but untruthful in the refusal case; the workspace path already shows the better pattern.
4. **Counting could use herdr's own numbers.** `TabInfo.pane_count` and `WorkspaceInfo.tab_count` are required snapshot fields; using them would replace three map walks with O(1) reads and remove any chance of drift between muxr's maps and herdr's counts. Related leftovers: `closeAgentRoute` counts panes and then `closeExactPane` counts them again, and the `paneCount === 0` branch (L269) is still unreachable.
5. **Carried over, unchanged:** the `agent-route-ambiguous` branch is dead in production (`IdentityStore` is keyed by `sessionId`), `identity.all()` copies every identity per Stop where `identity.get` would do, and the deleted one-line rationale for closing the status watch before `identity.remove` is still worth restoring.
6. **Physical-verification note:** when Stop closes a sole-pane tab, the workspace survives because another tab remains — often the shell tab herdr created with the workspace. Worth recording tab counts before/after in the evidence so a surviving bare shell tab is not read as a leak, and an empty-looking space is not read as an escalation.

Approved on substance for `39eb32de6f646f81c58a3ff8c2d8741b0ff69cfd`. Items 1-6 are polish, not gates.

PR #162, opened 2026-08-29
Fix physical agent presentation regression

umeranjum17:
## What

- restore responsive 300×200 terminal cards in the horizontal activity strip without changing lifecycle order, polling bounds, or the one-second full-visibility acknowledgement
- backfill canonical Agent Names in the enabled pane-titler plugin, parse OMP live chrome into Task Titles, and relink bundled plugins when their manifest version changes
- count unique live Agent Routes in Spaces while retaining stable shell/lifecycle terminal cards
- replace recolored placeholder provider marks with official light/dark assets and the distinct upstream OMP mark

## Why

Physical build 45 regressed cards to 208×152 and made their unchanged radius look excessively rounded. The reported `Agent` / `Untitled task` labels and missing OMP provider came from the running 0.1.20 host: live Herdr already exposed names and terminal titles, while 0.1.21 owns the canonical mapping. Mobile remains a consumer of backend names/titles and adds no invented fallback aliases.

## Verify

```bash
yarn run check
```

`34/34 passed in 122.6s`

Also verified focused activity/count flows, root and mobile typechecks, pane-titler backfill against live Herdr, host diagnostics reporting 0.1.21, and an x86_64 API 36 release APK installed and exercised against the live Herdr tree.


umeranjum17 (commented):
## REQUEST CHANGES — PR #162 @ `297d936d241c345ee3e86bbf7bdb7ed429855a79`

https://github.com/umeranjum17/muxr/pull/162 · base `main` @ `8cf72227` · 37 files, +214/−60 · review rebased from `d8ba5545` onto the amended head.

CI on this SHA: `suite` **pass** 4m52s ([run](https://github.com/umeranjum17/muxr/actions/runs/33234559990/job/99053140953)), `CodeQL` pass, `Analyze (actions)` pass, `Analyze (javascript-typescript)` pass.

The card restoration, the OMP Task-Title parse, the Agent Name backfill, the Spaces route counting and the icon replacement all verify. Four blockers, three of them introduced by this HEAD.

---

### Blockers

**B1 — a starting agent renders `Shell` on the LIVE strip.**
`apps/mobile/sources/herd/domain/herd.ts:104-108` labels every pane with no `agentKind` as `Shell`. `herdrTree()` derives that field from Herdr's agent registration alone — `...(agent?.agent === undefined ? {} : { agentKind: agent.agent })` (`apps/host/src/agent/infrastructure/herdrSessionSource.ts:1612`). `startSession` creates the pane with `tab.create`, adopts the route with its `kind`, transitions to `starting`, and only then fires `agent.start` followed by `waitForInteractiveAgent(paneId, 60_000)` (`herdrSessionSource.ts:979-1020`, `725-734`). For that whole window the tree carries `sessionId` with no `agentKind`, so every phone-launched agent shows `<Task Title> · Shell · Starting`, and a `start-timeout` / `start-launch-failed` launch keeps the label. `pluginPublicContext` already resolves this correctly — `agentKind: agent?.agent ?? session?.kind` (`herdrSessionSource.ts:1261`).
Fix: emit `agentKind: agent?.agent ?? session?.kind` from `herdrTree()`, and make the mobile rule `agentKind === undefined || agentKind === 'shell' ? 'Shell' : name`. `'shell'` is already the explicit shell kind (`createTab` adopts `kind: options.kind ?? 'shell'`; it is listed in `PROVIDER_KINDS`, `apps/host/src/agent/domain/identity.ts:11`).

**B2 — the Shell rule bypasses the single display vocabulary, so one pane carries two names on one screen.**
`liveTerminalIdentity` (`apps/mobile/sources/herd/application/liveTerminalOrder.ts:43`) is a second naming authority beside `agentPresentation.agentLabels`, whose doc comment is `/** One display vocabulary: Task Title, then canonical speakable Agent Name. */`. At this HEAD the same shell pane reads `Shell` in the strip but `Agent` in the Spaces row subtitle and its accessibility label (`SpacesTree.tsx:213`, `:234` → `agentLabels` → `paneDisplayName` → `normalizeAgentName(undefined)` = `'Agent'`), and `(Agent)` in the close-pane dialog (`SpacesTree.tsx:388-390`). The LIVE strip and SPACES render simultaneously on the Herd screen. Move the rule into `agentPresentation`/`paneDisplayName` (which needs `agentKind` in scope) and delete `liveTerminalIdentity`; strip, Spaces, dialog and a11y then agree by construction. Note `herdPanes` already writes `name: 'Shell'`, so `liveTerminalIdentity` re-decides what the producer decided — two sources of truth for one label.

**B3 — the README loses the OMP mark on GitHub's light theme.**
`docs/agents/icons/omp.svg` is the upstream mark verbatim (I diffed it byte-for-byte against `can1357/oh-my-pi@main:assets/icon.svg`, identical modulo comments), which fills the π with `#fafafa`. `README.md:169` embeds it at `width="28"` in a row of `#a3e635` glyphs. Rasterized over `#ffffff` only the 20×16 orange connector and two dots survive; over `#0d1117` it reads correctly. Its `120×90` viewBox also lands the badge at 28×21 among 28×28 siblings.
Fix it the way the mobile pair already does: a light variant using the mark's own `#0d0d0d` (measured — `omp-light.png` bars `grey5`, `omp-dark.png` bars `grey98`, `#f97316` preserved in both), or a `<picture>` with `prefers-color-scheme`.

**B4 — attribution no longer matches what ships.**
`apps/mobile/sources/assets/agents/ATTRIBUTION.md` now states that marks are unmodified LobeHub light/dark or colour variants and that "Unknown agents render their accessible text initial instead of an invented logo". `droid.png` and `maki.png` are still muxr-accent letter tiles (measured `#a3e635`, "D"/"M" in a rounded box) and both remain keys in `agentImages` (`AgentGlyph.tsx:94`, `:102`), so they render the invented tile and never the text initial. `docs/agents/icons/ATTRIBUTION.md` deleted the only sentence disclosing that its 20 sibling SVGs are accent recolors (`pi.svg` etc. are still `fill="#a3e635"`) while now describing assets that live in a different tree. Either drop `droid`/`maki` from `agentImages` so the promised text initial actually renders, or restore the placeholder/recolor disclosure in both files.

---

### Verified

**Card geometry — restored and lifecycle-independent.** On the running x86_64 API 36 emulator at 420 dpi (2.625 px/dp): card 1 spans x=42→829 px and card 2 starts at x=861 px → **300.0 dp wide, 12 dp gap**; top edge y=559, body/footer boundary y=957, bottom y=1083 → **199.6 dp tall**. Heading renders `LIVE`. `cardWidth = min(300, max(240, stripWidth − 40))` depends only on layout, never on lifecycle or status, and `reconcileLiveTerminalCards` now also compares `agentKind`, so a kind arriving late cannot churn card identity or reorder slots.

**Acknowledgement geometry — unchanged contract.** Same 1 s timer, same `measureInWindow`, same two gates: whole strip inside `[visibilityTop, windowHeight − bottomInset]` and whole card inside `[scrollX, scrollX + stripWidth]`. The spec was renumbered in place and every negative case still fails for its own reason (`stripTop: 79` on the top gate, `349` because 349+200 > 500, `scrollX: 20` because card 0 starts at 16). Only edge: the 240 dp floor exceeds the content box below ~280 dp of strip width, where nothing could ever ack — no shipping form factor reaches that.

**No polling or perf regression.** `TerminalPreview` keeps `POLL_MS = 3000`, `MAX_LINES = 24`, `paused = |index − firstVisible| > 2` and `scrollEventThrottle={100}`. Wider cards add no requests — the taller body just shows ~17 of the 24 lines already fetched instead of ~11. The only new per-render work is one `Math.min`/`Math.max`.

**Agent Name backfill — deterministic, route-safe, no numeric fallback.** `animalFor` is FNV-1a over `pane_id` then the first unused animal; `ensureAgentNameFrom` now writes `agent.name = fallback` into the local rows, so a multi-agent pass cannot collide; exhaustion returns `undefined` rather than suffixing; any displayable existing name short-circuits, which makes reruns no-ops — required, since setup invokes `backfill.mjs` on the version-matched path too. Renames key on the live `pane_id`; Agent Routes are untouched and names are never routing keys. Verified against live Herdr: `muxr.pane-titler 0.1.1` linked and enabled, `agent list` shows the animal `puma` on the unnamed `pi` pane with user-set names preserved elsewhere.

**Setup gate is safe.** `herdr 0.8.0` (= `MIN_HERDR`) does emit per-plugin `version` in `plugin list --json`, so the new required-field validation will not brick setup, and all 15 bundled manifests declare `version`. `agent list` is the same `cli:*` RPC as `plugin list`, so the backfill introduces no new daemon dependency for `ensureBundledPlugins`.

**OMP Task Titles — correct against real chrome.** Live `herdr agent list` right now yields `π ⠇ Supervise muxr release and merges`, `π ⠴ Fix mixed agent and shell pane labels`, `π > Fix architecture test boundary violation`; all three reduce correctly, and the strip order (`omp` prefix before the empty check) drops a bare `π` to `undefined`. Every chrome-titled record in `~/.muxr/host/herdr-identity.json` is `ours: false`, so `observedTaskTitle` re-derives it on the next observation — no migration needed. Two follow-ups, not blockers: `π > pockit` now yields the Task Title `pockit` (a cwd basename), and the identical `π - <cwd>` chrome from kind `pi` is untouched (`π - pockit`, `π - muxr-stop-host.X81m2p` are live).

**Spaces counts unique current Agent Routes.** `agentCount` = unique `sessionId` among panes with an `agentKind`; `buildSpaceRows` reads only the live tree, and `reconcileLiveTerminalCards` drops any card absent from the current selection, so retained strip slots cannot inflate the badge. The field only feeds the badge and its a11y string (`SpacesTree.tsx:286`, `:304`). The dedupe is defensive rather than load-bearing: `IdentityStore.byPane` matches one `paneId` per record, so a 0.1.21 tree cannot repeat a route.

**OMP is returned by backend kinds.** `agentKindsFromManifests` always prepends `['omp','mastracode']`, reaching mobile through `herdr.agentKinds` (dispatcher allowlist `createRequestDispatcher.ts:63`, handler `:128`).

**Old-host diagnosis is correct.** At `v0.1.20` the host source (`apps/host/src/herdr/herdrSessionSource.ts`) emits no `displayName` or `taskTitle` anywhere in the tree, and its `agentKinds()` returns raw `server.agent_manifests` with no hook-only injection — precisely the reported `Agent` / `Untitled task` / missing-OMP triple. The unit now runs `release-0-1-21/dist-npm/cli.mjs`.

**Icons are official and distinct where replaced.** `cursor`, `grok`, `hermes`, `opencode`, `pi` keep their LobeHub shapes with the `#a3e635` recolor removed — pure black light / pure white dark, no muxr accent anywhere. `omp` changes shape from the invented "O" tile to the upstream Oh My Pi π mark, distinct from `pi`'s pixel π.

**No service or runtime state committed.** 37 files, all source/assets/docs/scripts — no units, no `~/.muxr` state, no APKs or logs. Test growth respects `AGENTS.md`: assertions folded into the existing integration and lifecycle specs, zero new test files.

**Localization.** `liveTerminals.title` already existed unused in `_default.ts` and all 10 translations, so wiring it is a net removal of a hardcoded string.

### Nits

- `identity.ts:68` uses `kind?.toLocaleLowerCase()` for a protocol token; `toLowerCase()` is the locale-independent form (harmless for `omp`, but the same line-range mixes both).
- `herdr.mjs:208` passes a literal `false` for `dryRun` on a path where `dryRun` is already provably false — pass `dryRun` for symmetry with `:195`.
- `LiveTerminalsRow.tsx:143`: the `24` in `stripWidth − STRIP_GUTTER − 24` is unexplained; the content box is `stripWidth − 2 * STRIP_GUTTER`.
- The zero-state string next to the newly translated heading stays hardcoded English.

Re-request review after B1–B4; everything else is ready to merge.

umeranjum17 (commented):
## REQUEST CHANGES (one blocker) — PR #162 @ `99d933fae365f63cdfbd3395a76411037e900604`

Re-review rebased from `297d936d` onto `99d933fa`. base `main` @ `8cf72227`.
CI on this SHA: `suite` **pass** 4m44s ([run](https://github.com/umeranjum17/muxr/actions/runs/33236189852/job/99057478011)), `CodeQL` pass, `Analyze (actions)` pass, `Analyze (javascript-typescript)` pass.

All four previous blockers are genuinely closed — verified below, including provenance fetched from both upstreams. One new blocker fell out of the B1 fix: the host now publishes `agentKind: 'shell'`, and nothing treats `'shell'` as the shell sentinel.

---

### Blocker

**B5 — a phone-started Shell is labelled `Agent`, counted as an agent, and auto-expands its space.**

The B1 fix broadened the tree's kind source to the adopted session (`herdrSessionSource.ts:1603` → `:1613`), which is right for pending agents. But `'shell'` is itself an adopted kind: the Dock's first option is `shell` (`persistence.ts:17` — `AGENT_TYPES = ['shell', ...AGENT_KINDS]`), `StartAgentFromDock.providerKind` reaches `ops.ts:241` as `kind: options.agent`, so `session.start { kind: 'shell' }` lands at `herdrSessionSource.ts:899`, is adopted verbatim at `:980`, and takes the shell branch at `:1002`. From then on the pane travels the wire as `agentKind: 'shell'`, and `metadata.agentKind` has carried the same value all along (`infoFor` → `record.kind ?? agent?.agent`).

Nothing filters that value, so five branches misfire for exactly the case this PR set out to fix:

| Site | Result |
|---|---|
| `agentPresentation.ts:25-31` | `agentKind` defined → `agentName` = `metadata.agentName` = `'Agent'` (host always sends `displayName: normalizeAgentName(agent?.name)` = `Agent` for an agentless pane), never `Shell` |
| `herdTree.ts:60` | counted in `agentCount` → a space holding one phone-started shell reads "1 agent" |
| `herdTree.ts:15` `hasAgent` | true → `useHerdTreeLive.ts:57` auto-expands that space by default |
| `PaneGridView.tsx:192` | `Agent · shell` |
| `workspace/[id].tsx:102-104` | `Agent · shell` |

Two shells sitting side by side now disagree: the one opened in the terminal (no adopted kind) reads `Shell` and is uncounted; the one started from the phone reads `Agent` and counts as an agent. The count and auto-expand behaviours are also regressions against `main`, where `agentKind` could never be `'shell'`.

The invariant already exists one directory over — `homeDockEnvironment.ts:35` writes `...(key === 'shell' ? {} : { agentKind: key })`, i.e. shell is deliberately *not* an agent kind. Mirror it at the source and every consumer above is fixed without touching mobile:

```ts
// herdrSessionSource.ts:1603
const agentKind = agent?.agent ?? (session?.kind === 'shell' ? undefined : session?.kind);
```

Keep `taskTitleFor(undefined, agentKind)` on the raw kind if you still want `Shell task` titles. If you would rather let the kind through, then `agentLabels` needs `agentKind === 'shell'` folded into its `Shell` branch **and** `herdTree.ts:15`/`:60` need the same exclusion — three sites instead of one, which is why the host-side fix is the better one.

---

### Previous blockers — closed

**B1 (pending session kind) — fixed.** `herdrSessionSource.ts:1603` hoists `agent?.agent ?? session?.kind` and emits it at `:1613`. Traced end to end: `session.start {kind:'pi'}` adopts the kind at `:980`, `snapshotFor` returns immediately, `sessionInfoToSession` writes `metadata.agentKind` (`sessionMapping.ts:90`, schema `storageTypes.ts:179`), and the tree carries `'pi'` from the first poll — so neither the tree-driven nor the catalog-driven surface can read `Shell` during the up-to-60 s `agent.wait` window. `agentLabels` consults both (`pane?.agentKind ?? metadata?.agentKind`), and the new `pending` case in `liveTerminalOrder.spec.ts` pins the metadata-only path.

**B2 (single display vocabulary) — fixed, and better than asked.** `liveTerminalIdentity`, `paneDisplayName` and `paneTaskTitle` are deleted with zero remaining references anywhere in `apps/` or `packages/`; `utils/herdTree.ts` drops the re-exports. All six surfaces now resolve through `agentLabels`: LIVE card (`LiveTerminalsRow.tsx:90-95`), Spaces row and its a11y label (`SpacesTree.tsx:214`, `:237`), close dialog (`:389`), pane grid (`PaneGridView.tsx:155`), workspace grid (`workspace/[id].tsx:82`), terminal header (`TerminalScreen.tsx:366`). `agentName` is now a required `string`, which let the `?? ''` and `=== '' ? null` guards go. Moving `HERD_STATUS_LABELS` into `agentPresentation.ts:12-20` also breaks the old `herd ↔ agentPresentation` import cycle while the re-export at `herd.ts` keeps `@/utils/herd` consumers working. The double application in `LiveTerminalCard` (card fields already came from `agentLabels` inside `herdPanes`) is idempotent — `normalizeAgentName` is a fixed point on an already-normalized name.

**B3 (README OMP visibility) — fixed.** `omp-badge.svg` wraps the unchanged mark in `<g transform="translate(12 25) scale(.8666667)">` on a `#171717` rounded square: 120×90 → 104×78, leaving 12 px left/right and 25 px top/bottom — centred, uniformly scaled, no recolor (`#fafafa`, `#f97316`, `#0d0d0d` all intact). Rasterized at 28 px it reads cleanly over both `#ffffff` and `#0d1117`. `README.md:153` points at the badge, and the dropped marks are now a plain name list at `:156`.

**B4 (attribution matches what ships) — fixed.** The tree now holds exactly `omp.png`, `pi-light.png`, `pi-dark.png` and `ATTRIBUTION.md`; `AgentGlyph` requires exactly those three and nothing else — no dangling `require`. Both attribution files describe precisely that set. Provenance checked against the upstreams, not just the text:
- `docs/agents/icons/pi.svg` is `https://pi.dev/logo-auto.svg` verbatim — identical `<style>` block and identical path data, only the XML prolog and comments stripped.
- `docs/agents/icons/omp.svg` is still byte-identical to `can1357/oh-my-pi@main:assets/icon.svg` modulo comments.
- `omp.png` (128×128) is that mark on the dark badge; `pi-light.png`/`pi-dark.png` are 128×128 greyscale exports of the pi.dev geometry with correct polarity (black for light, white for dark). No `#a3e635` remains in any shipped asset.

### Carried forward from the earlier heads

Unchanged by this HEAD, so the earlier verification still stands: `CARD_WIDTH/HEIGHT/GAP = 300/200/12` and `STRIP_GUTTER = 16` are untouched, as is the 1 s acknowledgement effect and `visibleActivityEventIds`. Measured on the running x86_64 API 36 emulator at 420 dpi on the previous head: card 1 x=42→829 px, card 2 starts at 861 px (300.0 dp, 12 dp gap), card top y=559 / footer boundary 957 / bottom 1083 (199.6 dp), heading `LIVE`. I could not re-measure at this SHA — the emulator is no longer attached — but nothing in `297d936d..99d933fa` touches geometry, polling (`POLL_MS 3000`, `paused` window ±2, `scrollEventThrottle 100`) or the ack path. Also still good: pane-titler backfill determinism and route safety, the `plugin list --json` version gate against real `herdr 0.8.0`, OMP in `agentKindsFromManifests`, the `v0.1.20` diagnosis, unique-route counting, and no service or runtime state committed.

### Follow-ups (non-blocking)

- `docs/agents/icons/pi.svg` relies on an in-SVG `@media (prefers-color-scheme: dark)`. In an `` context browsers resolve that against the OS scheme, not the GitHub theme, so a reader on GitHub-dark + OS-light gets a black π on `#0d1117` (renderers that ignore the query — librsvg, for instance — fall back to `#000`, which is what my rasterization shows). It is the vendor's own adaptive asset used as intended, so this is not a licensing or authenticity problem; if you want the README bulletproof, either give Pi the same dark badge treatment as OMP or use `<picture>` with `media="(prefers-color-scheme: dark)"`, which resolves against the GitHub theme.
- `identity.ts:68` still uses `kind?.toLocaleLowerCase()` on a protocol token; `toLowerCase()` is the locale-independent form.
- `herdr.mjs:208` passes a literal `false` for `dryRun` where it is already provably false.
- `LiveTerminalsRow.tsx:143` — the `24` in `stripWidth − STRIP_GUTTER − 24` is still unexplained; the content box is `stripWidth − 2 * STRIP_GUTTER`.
- Task Titles: `π > pockit` still reduces to the cwd basename `pockit`, and the identical `π - <cwd>` chrome from kind `pi` is untouched.
- `liveTerminalOrder.ts:41-42` now has a stray double blank line where `liveTerminalIdentity` was removed.

One line in `herdrSessionSource.ts:1603` and this is ready to merge.

umeranjum17 (commented):
## APPROVE — PR #162 @ `02a8d0326647285ae10db46e8faf777c8b1c7690`

Third review, rebased from `99d933fa`. base `main` @ `8cf72227` · 74 files, +352/−193.
Checks on this SHA: `Analyze (actions)` SUCCESS, `Analyze (javascript-typescript)` SUCCESS, `CodeQL` SUCCESS. No `suite` row is registered for this SHA; I am taking the owner's `yarn run check` 34/34 as the suite evidence (the same command passed in CI on the two prior heads, `d8ba5545` 4m59s and `99d933fa` 4m44s, and the head-to-head delta since then is 6 files).

B5 is closed, and closed better than the one-liner I proposed. All five blockers are resolved. No further findings.

---

### B5 — fixed at every boundary, not just the tree

`publicAgentKind` (`herdrSessionSource.ts:73-75`) is applied at every point where a kind crosses into a payload or into storage. I enumerated all `agentKind` emissions in `apps/host/src` and each one is covered:

| Site | Path |
|---|---|
| `:423` → `:442` | `infoFor` → `SessionInfo.agentKind` → mobile `metadata.agentKind` |
| `:1609` → `:1619` | `herdr.tree` panes |
| `:1242` | plugin public context sessions |
| `:1260` → `:1266` | plugin public context tab panes |
| `:1420` → `:1425` | realtime plugin public context |
| `:990` | `startSession` → `identity.adopt` |
| `:1919` | `createTab` → `identity.adopt` |

`pluginPublicContext.ts:84`/`:131` are pass-through sanitizers over those values, and `agentKinds()` is the installable catalog, which never contained `shell`. Normalizing at **both** the adopt sites and every read site is the right call rather than redundancy: `parseAgentIdentity` restores `kind` verbatim from `herdr-identity.json`, so a record persisted by an intermediate build could still hold `'shell'` on disk — the read-side wrapping means such a record can never leak to a client.

Control flow that legitimately still needs the literal is untouched: `startSession`'s `kind === 'shell'` branch at `:1006` reads the local raw variable, `createTab` still passes `options.kind` to `agent.start` (`:1923`, `:1928`), and `paneSplit` keeps `splitOptions.kind` (`:1671`, `:1675`). `startSession` computes `taskTitle` from the raw kind at `:906`, so a dock-started Shell keeps its `Shell task` title.

Consequently all five mobile branches I listed now behave, with no mobile change required: `agentPresentation.ts:25-31` yields `Shell`, `herdTree.ts:60` excludes it from `agentCount`, `herdTree.ts:15` → `useHerdTreeLive.ts:57` no longer auto-expands a shell-only space, and `PaneGridView.tsx:192` / `workspace/[id].tsx:102-104` render `Shell` instead of `Agent · shell`.

Two reads intentionally keep their generic fallback — `realtimeAgentFor.kind = record.kind ?? 'agent'` (`:1050`) and the layout snapshot's per-pane kind (`:1793`). Both now report the same thing for a dock-started shell as they already did for a terminal-opened one, so this is alignment rather than loss.

**E2E** (`checkHerdrE2E.mjs:155-171`) covers both delivery paths, which is what makes the claim checkable: it asserts `phoneShell.info.agentKind === undefined` (the `SessionInfo`/metadata path) and that the matching `herdr.tree` pane also omits it, then compares against a pre-existing kind-less session captured from `session.list` at `:148` before the shell is created, so it cannot self-select. The terminal-opened half is conditional on such a pane existing; on a normal tree it does (13 of 27 identity records on my host carry no kind). Cleanup is sound: both starts use the same `workdir`, and `startSession` resolves one workspace per directory by `label === cwd` (`:947-956`), so overwriting `createdWorkspaceId` at `:158` and `:176` still names the single workspace that `finish()` closes, and the shell session is explicitly stopped at `:170`.

### Pi README badge — verified

`pi-badge.svg` is `#fafafa` rounded square + nested `<svg x="16" y="16" width="96" height="96" viewBox="0 0 800 800">` holding the mark. I diffed the path data against `https://pi.dev/logo-auto.svg`: both `d` attributes are character-identical after whitespace normalization, and the only fills are `#fafafa` (badge) plus `#000` ×2 — upstream's own light-mode colour, so the official mark is unchanged and unrecoloured. Rasterized at 28 px it is fully legible over both `#ffffff` and `#0d1117`, with no dependence on `prefers-color-scheme`, which resolves the last README risk I raised. `ATTRIBUTION.md` states exactly this. The nested-`<svg>` form renders correctly under librsvg.

The two badges are now inverse polarity side by side (white square for Pi, near-black for OMP). Both are deterministic and legible, so this is taste, not a defect — mentioning it only in case you want a matched pair.

### Also confirmed at this SHA

- Stray double blank line at `liveTerminalOrder.ts:41` removed.
- Asset tree is exactly `omp.png`, `pi-light.png`, `pi-dark.png` + `ATTRIBUTION.md`, and `docs/agents/icons` is exactly `omp.svg`, `omp-badge.svg`, `pi.svg`, `pi-badge.svg` + `ATTRIBUTION.md`. `AgentGlyph` requires precisely the three PNGs — no dangling `require`, no `#a3e635` anywhere.
- 74 files, every one source, script, doc or asset. No service units, no `~/.muxr` state, no APKs, no logs.
- Untouched since the heads where I verified them, so the earlier evidence carries: `CARD_WIDTH/HEIGHT/GAP = 300/200/12` measured on-device at 420 dpi (card 1 x=42→829, card 2 at 861; top 559, footer boundary 957, bottom 1083 → 300.0 × 199.6 dp) with the `LIVE` heading; the 1 s acknowledgement effect and `visibleActivityEventIds` gates; `POLL_MS 3000` / `paused` ±2 / `scrollEventThrottle 100`; pane-titler backfill determinism and route safety; the `plugin list --json` version gate against real `herdr 0.8.0`; OMP in `agentKindsFromManifests`; unique-route counting; the `v0.1.20` diagnosis; and `agentLabels` as the sole display vocabulary with `liveTerminalIdentity`/`paneDisplayName`/`paneTaskTitle` gone and zero remaining references.

### Optional follow-ups — none blocking

- `createTab` with neither a label nor a kind now titles the pane `Coding task` rather than `Shell task`, because `identity.adopt` computes `taskTitleFor(undefined, undefined)`. It matches what every terminal-opened shell already shows, so it is consistent; a `genericTaskTitle` shell case would be nicer copy.
- Mobile trusts the host to normalize. That is the documented layering, and no released host emits `agentKind: 'shell'`, so no guard is needed — worth a line in `agentLabels` only if you ever want old-host tolerance.
- Carried over: `identity.ts:68` `toLocaleLowerCase` on a protocol token; `herdr.mjs:208` literal `false` for `dryRun`; the unexplained `24` in `LiveTerminalsRow.tsx:143`; `π > pockit` still reducing to the cwd basename; `π - <cwd>` chrome from kind `pi` still unparsed.

Ship it.

umeranjum17 (commented):
## APPROVE (rebase delta) — PR #162 @ `536f357bfc9c0a9f0a56e8af0bbd3dea12f35820`

Rebased onto `origin/main` @ `85b22943` (#161, "close only the selected agent layout"). Merge base confirms `85b22943`; state `MERGEABLE`.
Checks on this SHA: `suite` **pass** 5m11s ([run](https://github.com/umeranjum17/muxr/actions/runs/33237433722/job/99060765116)), `Analyze (actions)` SUCCESS, `Analyze (javascript-typescript)` SUCCESS, `CodeQL` SUCCESS.

I diffed the PR's own content before and after the rebase — `git diff 8cf72227..02a8d032` against `git diff 85b22943..536f357b`, normalized for hunk headers — and the replay is byte-identical except **two** deltas. Both are correct. No new findings; the previous approval stands.

### Delta 1 — SpacesTree close copy: main's semantics preserved exactly

Both alert strings at `536f357b` are character-identical to `85b22943`:

- `Close workspace?` → `Closes only the "${name}" workspace in herdr. If that would close its worktree group, nothing closes.`
- `Close pane?` → `Closes only the pane for "${labels.taskTitle}"${identity} in herdr. If that would also close its tab, nothing closes.`

The only PR-side change is `const identity = \` (${labels.agentName})\`` in place of main's `labels.agentName === undefined ? '' : …`. That is behaviour-preserving here, not a shortcut: this PR makes `AgentLabels.agentName` a required `string` (`agentPresentation.ts`), so the undefined arm was unreachable. Shared Agent/Shell labelling is retained, pane-only wording and tab-safety wording are untouched.

### Delta 2 — the E2E's shell `session.stop` is gone, and it had to be

`await request(socket, 'session.stop', { sessionId: shellId })` was dropped from `checkHerdrE2E.mjs`. This was outside the declared conflict, so I traced it against #161's new guard rather than assume it was cosmetic — it is *required*:

The E2E's `workdir` is a fresh `mkdtemp`, so the shell start creates the workspace (`workspace.create`, `label === cwd`) with exactly one tab holding one pane. `closeAgentRoute` counts panes in the route's tab, finds one, and delegates to `closeExactTab`, which throws `tab-close-would-widen` when its workspace has `tabCount <= 1`. So the previously-approved line would now **fail** the E2E rather than clean up. Removing it is the right resolution.

One consequence worth a comment, not a change: the surviving `session.stop({ sessionId: newId })` at `:223` now passes *because* the shell tab is still open — the workspace has two tabs, so closing the pi tab no longer widens. The block ordering and the absence of shell cleanup are therefore load-bearing. A one-liner at the removal site ("leave the shell tab open: the pi stop below would widen to the workspace otherwise") would stop someone reinstating cleanup or reordering the blocks and hitting a confusing `tab-close-would-widen`. Nothing after `:223` is affected — it is the last mutation before `finish(0, …)`.

### Main's stop-scope code is untouched — verified, not assumed

`git diff 85b22943..536f357b` over `stopAgent.ts`, `catalog/application/ops.ts`, `control-plane/domain/requests.ts`, `terminalManager.ts` and `openTerminal.ts` is **0 lines**. The `herdrSessionSource.ts` delta contains no `closeExactPane` / `closeExactTab` / `closeAgentRoute` / `would-widen` / `async stop` line. #161's artefacts all survive intact:

- `TerminalScreen.tsx` keeps `import { sessionStop } from '@/catalog/ops'` (`:21`), the `void sessionStop(props.id)` call (`:333`), and its full "Closes only this agent pane, or its tab when the pane is alone…" copy (`:326`). This PR's delta on that file is only the `paneDisplayName` → `agentLabels` swap.
- `layoutSelfCheck.ts` — this PR's delta versus the new base is exactly the one `parseTaskTitle('π ⠼ …', 'omp')` assertion; #161's stop self-checks are entirely intact.

### Everything else replays unchanged

The remaining 49 non-binary paths carry the content already reviewed at `02a8d032`: `publicAgentKind` at all seven kind boundaries plus both adopt sites, `agentLabels` as the sole display vocabulary, the 300×200 dp strip with its acknowledgement gates and unchanged polling constants, the pane-titler backfill, the `pi-badge.svg` / `omp-badge.svg` pair with upstream-identical marks, and the trimmed asset tree. Still no service units, runtime state, APKs or logs.

Ship it.

PR #163, opened 2026-08-29
fix(voice): prevent Gemini playback overflow and persist wake standby

umeranjum17:
## Problem
- Gemini Live can deliver multi-second PCM bursts faster than the phone's bounded native sink drains, producing `Realtime playback buffer overflowed`.
- Wake on speech expired after 30 minutes and several non-toggle paths cleared its persisted preference.

## Fix
- Pace Gemini PCM as ordered 20 ms frames behind a bounded queue; pause the provider WebSocket for stdout/mobile backpressure.
- Preserve immediate interrupt clear and explicit-stop teardown; genuine queue overload remains visible and recoverable.
- Remove standby TTL/expiry, persist the preference across restart/reconnect/background, and disable it only through the settings toggle.
- Update the settings copy and add provider-shaped burst, pacing, clear, >30-minute, service-order, and persisted-restart assertions.

## Verification
- Focused mobile flows: 12 passed.
- Voice/provider flows: 10 passed.
- Root and mobile typechecks passed.
- Full `yarn check` passed.
- API 36 x86_64 release APK built, installed, and cold-launched; background/foreground after a simulated >30-minute clock jump hot-resumed.
- Physical audible playback was not re-run: no physical Android device was attached; deterministic native admission/interrupt assertions passed. No merge or publish performed.

### DIFF
--- a/apps/mobile/modules/voice-overlay/android/src/main/java/expo/modules/voiceoverlay/VoiceOverlayModule.kt
+++ b/apps/mobile/modules/voice-overlay/android/src/main/java/expo/modules/voiceoverlay/VoiceOverlayModule.kt
@@ -123,6 +123,10 @@ class VoiceOverlayModule : Module() {
     Function("isRealtimePcmDrained") { realtimePcm.isDrained() }
     Function("stopRealtimePcm") { realtimePcm.stop() }
     Function("isServiceReady") { VoiceOverlayService.isVoiceForegroundReady() }
+    Function("setNetworkActive") { active: Boolean ->
+      VoiceOverlayService.setNetworkActive(active)
+      true
+    }
 
     // The whole Android audio session, so nothing else gets to argue about it.
     // Communication mode is what engages the hardware echo canceller; WebRTC
--- a/apps/mobile/modules/voice-overlay/android/src/main/java/expo/modules/voiceoverlay/VoiceOverlayService.kt
+++ b/apps/mobile/modules/voice-overlay/android/src/main/java/expo/modules/voiceoverlay/VoiceOverlayService.kt
@@ -38,6 +38,11 @@ class VoiceOverlayService : Service() {
     private const val HERD_GROUP_KEY = "muxr.herd"
     @Volatile private var voiceForegroundReady = false
     internal fun isVoiceForegroundReady(): Boolean = voiceForegroundReady
+    @Volatile private var voiceNetworkActive = true
+    internal fun setNetworkActive(active: Boolean) {
+      voiceNetworkActive = active
+      instance?.get()?.applyNetworkLock(active)
+    }
 
     private var instance: WeakReference<VoiceOverlayService>? = null
     private var herdMode = "offline"
@@ -59,9 +64,8 @@ class VoiceOverlayService : Service() {
     private var voiceMuted = false
     private var voiceStartedAt = 0L
     /**
-     * Held only while a voice session is live: keeps the CPU and the Wi-Fi
-     * radio responsive with the screen off so the realtime socket and PCM
-     * playback do not stall. Both are released on stop/destroy/timeout.
+     * The CPU lock keeps background microphone detection alive. The Wi-Fi
+     * low-latency lock is held separately only for connected realtime traffic.
      */
     private var voiceWakeLock: PowerManager.WakeLock? = null
     private var voiceWifiLock: WifiManager.WifiLock? = null
@@ -535,6 +539,14 @@ class VoiceOverlayService : Service() {
         ?.apply { setReferenceCounted(false) }
     }
     runCatching { if (voiceWakeLock?.isHeld == false) voiceWakeLock?.acquire() }
+    applyNetworkLock(voiceNetworkActive)
+  }
+
+  private fun applyNetworkLock(active: Boolean) {
+    if (!active) {
+      runCatching { if (voiceWifiLock?.isHeld == true) voiceWifiLock?.release() }
+      return
+    }
     if (voiceWifiLock == null) {
       val wifi = applicationContext.getSystemService(Context.WIFI_SERVICE) as? WifiManager
       voiceWifiLock = wifi?.createWifiLock(WifiManager.WIFI_MODE_FULL_LOW_LATENCY, "muxr:voice")
--- a/apps/mobile/modules/voice-overlay/index.ts
+++ b/apps/mobile/modules/voice-overlay/index.ts
@@ -24,6 +24,7 @@ interface VoiceNative {
     stopRealtimePcm: () => RealtimePcmStats;
     startService: () => boolean;
     isServiceReady: () => boolean;
+    setNetworkActive: (active: boolean) => boolean;
     stopService: () => boolean;
     startHerdService: () => boolean;
     stopHerdService: () => boolean;
@@ -69,6 +70,15 @@ export function isVoiceServiceReady(): boolean {
         return native?.isServiceReady() ?? false;
     } catch {
         return false;
+    }
+}
+
+/** Wi-Fi latency is needed only while realtime provider traffic is active. */
+export function setVoiceNetworkActive(active: boolean): void {
+    try {
+        native?.setNetworkActive(active);
+    } catch {
+        // The CPU/microphone foreground service remains the standby fallback.
     }
 }
 
--- a/apps/mobile/modules/voice-overlay/ios/VoiceOverlayModule.swift
+++ b/apps/mobile/modules/voice-overlay/ios/VoiceOverlayModule.swift
@@ -51,6 +51,7 @@ public final class VoiceOverlayModule: Module {
     // iOS has no Android-style foreground service. These keep the shared JS
     // lifecycle platform-blind while APNs owns background agent notifications.
     Function("startService") { true }
+    Function("setNetworkActive") { (_: Bool) -> Bool in true }
     Function("stopService") { true }
     Function("startHerdService") { false }
     Function("stopHerdService") { true }
--- a/apps/mobile/sources/app/(app)/settings/voice.tsx
+++ b/apps/mobile/sources/app/(app)/settings/voice.tsx
@@ -136,7 +136,7 @@ export default function VoiceProviderScreen() {
                     />
                 </ItemGroup>
             )}
-            <ItemGroup title="Hands-free" footer="Audio stays on this device until speech is detected. Standby turns itself off after 30 minutes.">
+            <ItemGroup title="Hands-free" footer="Listens only on this device until speech is detected. The setting stays enabled until you disable it and uses additional battery.">
                 <Item
                     title="Wake on speech"
                     subtitle="Reconnect realtime voice when someone starts talking"
--- a/apps/mobile/sources/catalog/application/localSettings.ts
+++ b/apps/mobile/sources/catalog/application/localSettings.ts
@@ -14,7 +14,7 @@ export const LocalSettingsSchema = z.object({
     zenMode: z.boolean().describe('Hide all sidebars and non-essential UI for focused work'),
     promotedNotificationsPrompted: z.boolean().describe('Whether Android Live Updates access was already explained'),
     backgroundConnectionPrompted: z.boolean().describe('Whether Android background activity settings were already explained'),
-    vadStandbyEnabled: z.boolean().describe('Wake realtime voice from bounded local speech activity standby'),
+    vadStandbyEnabled: z.boolean().describe('Persistently wake realtime voice from local speech activity standby'),
     // Herd tab: bucket the agents section under workspace subheaders (herdr's "grouped" toggle).
     // Saved herdr tab layouts (split tree + agent kind per pane), newest first.
     savedLayouts: z
--- a/apps/mobile/sources/catalog/application/sessionSync.integration.spec.ts
+++ b/apps/mobile/sources/catalog/application/sessionSync.integration.spec.ts
@@ -564,11 +564,15 @@ describe('session sync flow', () => {
             };
         }
         mmkvValues.set('lifecycle-voice-reports-v1', JSON.stringify(persistedVoice));
+        state.applyLocalSettings({ vadStandbyEnabled: true });
 
         // Module re-evaluation simulates the store/app restarting while MMKV remains.
         vi.resetModules();
         const restarted = (await import('./storage')).storage;
         restarted.getState().setLifecycleScope('test-authority:machine');
+        expect(restarted.getState().localSettings.vadStandbyEnabled).toBe(true);
+        restarted.getState().applyLocalSettings({ vadStandbyEnabled: false });
+        expect(JSON.parse(mmkvValues.get('local-settings')!).vadStandbyEnabled).toBe(false);
         expect(restarted.getState().voicePendingReports).toEqual([durableReport]);
         expect(restarted.getState().voiceDeliveredReportIds).not.toContain('voice-collision');
         restarted.getState().setLifecycleScope('x'.repeat(201));
--- a/apps/mobile/sources/conversation/application/realtimeSession.spec.ts
+++ b/apps/mobile/sources/conversation/application/realtimeSession.spec.ts
@@ -1,5 +1,6 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import type { RealtimeWebRtcCallbacks } from '../infrastructure/realtimeWebRtc';
+import type * as VadStandbyModule from './vadStandby';
 
 const mocks = vi.hoisted(() => ({
     openStream: vi.fn(),
@@ -25,6 +26,7 @@ const mocks = vi.hoisted(() => ({
         routeVoiceAudio: vi.fn(() => true),
         releaseVoiceAudio: vi.fn(),
         startVoiceService: vi.fn(() => true),
+        setVoiceNetworkActive: vi.fn(),
         stopVoiceService: vi.fn(),
     },
     webRtc: {
@@ -224,7 +226,7 @@ describe('generic realtime stream session', () => {
         await new Promise((resolve) => setTimeout(resolve, 30));
         expect(mocks.pcm.playRealtimePcm.mock.calls.filter(([data]) => data === 'c3RhbGUh')).toHaveLength(1);
 
-        const burst = Array.from({ length: 4 }, (_, index) => Buffer.alloc(20_000, index + 1).toString('base64'));
+        const burst = Array.from({ length: 180 }, (_, index) => Buffer.alloc(960, index + 1).toString('base64'));
         const pausesBefore = stream.send.mock.calls.filter(([frame]) => frame.action === 'pause_output').length;
         const resumesBefore = stream.send.mock.calls.filter(([frame]) => frame.action === 'resume_output').length;
         mocks.pcm.playRealtimePcm.mockReturnValue(false);
@@ -396,10 +398,18 @@ describe('generic realtime stream session', () => {
         expect(stream.close).toHaveBeenCalledOnce();
     });
 
-    it('arms locally after two speech-energy chunks and keeps silence off the provider path', async () => {
-        const vad = await vi.importActual<typeof import('./vadStandby')>('./vadStandby');
+    it('arms locally without an expiry and keeps silence off the provider path', async () => {
+        vi.useFakeTimers();
+        const timeout = vi.spyOn(globalThis, 'setTimeout');
+        const vad = await vi.importActual<typeof VadStandbyModule>('./vadStandby');
         const wake = vi.fn();
-        await expect(vad.startVadStandby(wake, vi.fn())).resolves.toBe(true);
+        await expect(vad.startVadStandby(wake)).resolves.toBe(true);
+        expect(timeout).not.toHaveBeenCalledWith(expect.any(Function), 30 * 60_000);
+        expect(mocks.pcm.startVoiceService.mock.invocationCallOrder[0])
+            .toBeLessThan(mocks.liveAudio.init.mock.invocationCallOrder[0]!);
+        expect(mocks.pcm.setVoiceNetworkActive).toHaveBeenCalledWith(false);
+        await vi.advanceTimersByTimeAsync(31 * 60_000);
+        expect(vad.vadStandbyOwnsMicrophone()).toBe(true);
         const inspect = mocks.liveAudio.on.mock.calls.at(-1)![1] as (data: string) => void;
         inspect(Buffer.alloc(4_800).toString('base64'));
         expect(wake).not.toHaveBeenCalled();
@@ -410,16 +420,18 @@ describe('generic realtime stream session', () => {
         inspect(speech.toString('base64'));
         expect(wake).toHaveBeenCalledOnce();
         vad.stopVadStandby();
+        expect(vad.vadStandbyOwnsMicrophone()).toBe(false);
+        vi.useRealTimers();
 
         let finishInit!: () => void;
         mocks.liveAudio.init.mockImplementationOnce(() => new Promise<boolean>((resolve) => {
             finishInit = () => resolve(true);
         }));
         const startsBeforeCancellation = mocks.liveAudio.start.mock.calls.length;
-        const arming = vad.startVadStandby(vi.fn(), vi.fn());
+        const arming = vad.startVadStandby(vi.fn());
         await vi.waitFor(() => expect(finishInit).toBeTypeOf('function'));
         vad.cancelVadStandbyStart();
-        const newerArming = vad.startVadStandby(vi.fn(), vi.fn());
+        const newerArming = vad.startVadStandby(vi.fn());
         finishInit();
         await expect(arming).resolves.toBe(false);
         await expect(newerArming).resolves.toBe(true);
--- a/apps/mobile/sources/conversation/application/realtimeSessionState.ts
+++ b/apps/mobile/sources/conversation/application/realtimeSessionState.ts
@@ -4,6 +4,7 @@ import { sync } from '@/catalog/sync';
 import { getCachedConnectionSettings } from '@/connection';
 import {
     addVoiceNotificationActionListener,
+    setVoiceNetworkActive,
     startVoiceService,
     stopVoiceService,
 } from '@/../modules/voice-overlay';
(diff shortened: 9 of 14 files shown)


umeranjum17 (commented):
## REQUEST CHANGES — reviewed at `3746a314b982fb5a6254db3e053b95798a569da9` (base `8cf722275ad75e524cfcbc44b041357996c80281`, merge-base = base, clean worktree, read-only)

The Gemini pacing design is the right root fix and I want it in. Three cheap, concrete items block: one of the PR's own verification claims is demonstrably false, one teardown path can stall while holding a live mic, and the TTL removal silently unbounds a wakelock. Nothing here is a rewrite.

**Verified sound.** 20 ms framing is exact: `AUDIO_CHUNK_SIZE = 24000*2*20/1000*4/3` = 1280 base64 chars → 960 B PCM (`stream.mjs:46-47`), 4-aligned so every frame decodes independently, and it matches the native writer's `WRITE_SLICE_MS = 20` (`RealtimePcmPlayer.kt:15`). The burst now lands in the host's bounded 4 MB queue instead of the phone's 192 KB JS queue (`realtimePlayback.ts:7`) — that is the actual cause of `Realtime playback buffer overflowed`. FIFO single-consumer queue ⇒ no reordering; the only early exit in `emitAudio` is the overflow close, so no silent drops; 4 MB overflow surfaces an explicit `close('Voice output buffer overflowed.', true)` ⇒ genuine overload stays visible. `ws.pause()` on non-empty queue / blocked stdout / client pause is real provider backpressure. Moving `action:'stop'` ahead of the `ws.readyState !== OPEN` guard (`stream.mjs:282-299`) is a strict improvement — stop now works with the socket closed or still connecting, clears the queue, then force-closes so `realtime.closed` lands at head. xAI/OpenAI/Codex `stream.mjs` have **zero diff**; `pause_output`/`resume_output` predate this PR (`realtimeStream.ts:40,116`), xAI already consumed them, OpenAI/Codex still no-op them. TTL removal is total: zero hits repo-wide for `VAD_STANDBY_MS`, `expireVadStandby`, `resetExpiry`, `expiryCallback`, `onExpire`; the module-level `AppState` expiry hook is gone. FGS ordering is correct (`vadStandby.ts:58` `startVoiceService` before `LiveAudioStream.init/start` at `:70-84`); `foregroundServiceType microphone|dataSync` + `RECORD_AUDIO` declared. Privacy gate intact — nothing leaves the device before two loud frames win. Tests: 22 passed / 0 failed across the three touched specs; root typecheck clean. (First run showed 4 failures from a **stale hoisted `@muxr/contract` dist**, not this PR — `normalizeAgentName is not a function`; clean after rebuilding contract.)

### Blocker 1 — MAJOR: the ">30 minute" assertion is tautological; it passes against the unfixed code

`realtimeSession.spec.ts:405-411` installs fake timers *after* `startVadStandby()` has already resolved, so the pre-fix real `setTimeout(expireVadStandby, 30*60_000)` is never on the fake clock and `advanceTimersByTimeAsync` cannot fire it. The mocked `AppState.addEventListener` is never invoked either, so the second expiry path is untested too.

Proven, not argued: I reverted only `vadStandby.ts` + `realtimeSessionState.ts` to base-SHA contents, kept this PR's spec, and ran it — **1 passed**. The 31-minute advance did not expire standby on the broken code. Control probe: spying `globalThis.setTimeout` shows the pre-fix arm still schedules `delay === 1800000`, and moving `vi.useFakeTimers()` *before* `startVadStandby` correctly flips `vadStandbyOwnsMicrophone()` to `false`.

So the behaviour is genuinely fixed, but the guard offered as proof guards nothing, and the PR body's "simulated >30-minute clock jump" verification does not hold. Fix is a two-line reorder: hoist `vi.useFakeTimers()` above `startVadStandby`, keep the rest. Then it fails on base and passes here.

### Blocker 2 — MAJOR: non-forced `close()` is fully gated by the client pause, and can strand teardown for 10 minutes

`flushOutput` (`stream.mjs:60-71`) refuses any non-`force` head item while `outputPausedByClient`, and `close()` defaults `force = false`. Every close except the explicit client stop and the overflow path is non-forced: `finishTurn` `:195`, provider error `:274`, reconnect exhaustion `:344`, refusal `:379`, terminal `:388`, `main` catch `:414`.

Queuing `realtime.closed` behind pending audio is *correct* — the farewell should finish playing. Being blocked outright by `pause_output` is not. The reachable path: `RealtimePcmPlayer.failLocked` sets `playback = null` (`:278-284`), after which `write()` returns `false` forever (`:96`), so the mobile queue never drains, `outputPressured` never clears (`realtimePlayback.ts:204-212`), and `resume_output` is never sent. A model-initiated hangup or provider error then never reaches the host. Only bound is `pluginStreamManager.ts:32` `STREAM_IDLE_TIMEOUT_MS = 10*60_000` → SIGTERM, SIGKILL +2 s. Ten minutes of a stranded child holding a microphone FGS is not an acceptable floor for an ordinary hangup.

`realtime.closed` is terminal and tiny — pass `force = true` from `close()` unconditionally (or at least on the `finishTurn`/error paths), exactly as the stop path already does.

### Blocker 3 — MAJOR: TTL removal silently unbounds a partial wakelock + low-latency Wi-Fi lock; copy does not disclose it

`startVoiceService` → `ACTION_START` (`VoiceOverlayModule.kt:161`) → `acquireVoiceLocks()` (`VoiceOverlayService.kt:488`), which takes `PARTIAL_WAKE_LOCK` **and** `WIFI_MODE_FULL_LOW_LATENCY` with `setReferenceCounted(false)` (`:531-543`). The 30-minute TTL was the only thing bounding those; standby now holds CPU awake, Wi-Fi power-save disabled, and the mic FGS up indefinitely. During standby no audio leaves the device (that is the feature's whole premise), so the low-latency Wi-Fi lock in particular is holding a battery-expensive radio mode for a state that transmits nothing.

New copy — "Audio stays on this device until speech is detected. Wake on speech stays enabled until you turn it off." (`voice.tsx:139`) — is accurate but silent on an always-on mic plus indefinite CPU/radio locks. Either scope the Wi-Fi lock out of the standby-only case, or state the always-on cost in the footer. This is the tradeoff the TTL was hiding; make it explicit rather than incidental.

### Non-blocking

- **MINOR — forced frames only bypass the pause at queue head.** `flushOutput` inspects `outputQueue[0]` only, so the forced interrupt clear (`:213-216`) sits behind a non-forced frame — reachably `state('thinking')` emitted at `:222` before the turn's first audio — while paused. `clearQueuedAudio` removes audio only, so it does not clear that head. Self-heals in ≤4 s (mobile drains 192 KB, and `clear()` sends `resume_output` at `realtimePlayback.ts:246-247`), so barge-in lags rather than breaks. Scan for the first forced item, or drop non-audio frames too.
- **MINOR — pacing drift is monotonic and never corrected.** `setTimeout(flushOutput, item.delayMs)` is a fixed per-frame delay, not a deadline. Measured on this box: 400 frames carrying 8000 ms of audio took 8050-8055 ms wall — **+0.6 %**, and it never catches up within a turn. Native cushion is `PREBUFFER_MS = 250` (`RealtimePcmPlayer.kt:13`), so ~40 s of continuous speech when idle, but far less on a host that is by definition running coding agents; past that, `rearmLocked(midTurn=true)` pauses/flushes AudioTrack → audible gap, counted in `underruns`/`drainRestarts`. A deadline-based pacer (`next += 20; setTimeout(next - now)`) is a one-liner and removes the class. **Physical audible playback was not re-run in this review** (no device attached), so choppiness cannot be excluded from CI alone — this is the item most worth a device pass.
- **MINOR — arm failure leaves the toggle ON but dead.** Dropping `if (!armed) applyLocalSettings({vadStandbyEnabled:false})` means the `Switch` (`voice.tsx:145`, bound to the persisted pref) reads ON while nothing is armed. Repair only fires on foreground/auth/sessionCount/pref change (`KernelNotifications.tsx:119-122`) — socket status is not a dependency, so a host reconnect alone does not re-arm. Consistent with "only the toggle disables it", but the switch should reflect armed state, not intent.
- **MINOR — explicit stop no longer fully stops, for every provider.** `stopRealtimeSession` (`realtimeSessionState.ts:450-454`) sleeps and re-arms when the pref is true, so the notification/hangup stop leaves mic + FGS hot with no full-stop path outside settings. Provider-agnostic, so xAI/OpenAI/Codex inherit it. The transport itself does close correctly (`realtimeSession.ts:60-100`). Intentional, but confirm the notification "stop" affordance is meant to leave the mic live.
- **NIT — the pacer itself has no coverage.** Only pure `chunkAudio` is tested; `flushOutput`, the 4 MB overflow close, `ws.pause()/resume()`, and the forced clear/stop bypass are untested. Also `providerRefusal.spec.mjs:42-47` fills the burst with a constant byte (`Buffer.alloc(..., 7)`), so `frames.join('') === burst` cannot detect reordering — the real ordering guarantee comes from `realtimeSession.spec.ts:227,246` (`Buffer.alloc(960, index+1)`, `admitted === burst`), which is a genuine no-loss/in-order check. `providerRefusal.spec.mjs` already stands up a `WebSocketServer`, so one flow-level test driving a burst through the plugin's stdout would fit the repo's few-big-tests rule without a matrix.

CI at this SHA: suite, CodeQL, both Analyze jobs SUCCESS; mergeable CLEAN. Not re-run: physical audible playback, and the APK install/cold-launch claimed in the PR body.

Happy to flip to approve once Blockers 1-3 are addressed — the pacing architecture itself needs no rework.

umeranjum17 (commented):
## REQUEST CHANGES — re-reviewed at `af95fbe8fe76e722bd424819c908eb4964387f85` (base `8cf722275ad75e524cfcbc44b041357996c80281`)

All three blockers from `3746a314` are genuinely fixed, and I verified each by falsification rather than by reading the diff. One **new** blocker was introduced by the fix for my earlier arm-failure MINOR: a non-toggle caller can now silently persist `vadStandbyEnabled: false`, which contradicts this PR's headline requirement. One-line fix location given below; everything else is ready.

### Blocker 1 (>30 min guard) — FIXED, and now a real guard

`vi.useFakeTimers()` precedes `importActual` and `startVadStandby`, plus a direct `setTimeout` spy and a positive control. I re-ran the falsification: base-SHA `vadStandby.ts` + `realtimeSessionState.ts`, HEAD spec kept, and it **fails** — the opposite of last time:

```
realtimeSession.spec.ts:407
AssertionError: expected "setTimeout" to not be called with arguments: [ Any<Function>, 1800000 ]
Received:  1st setTimeout call: [ [Function expireVadStandby], 1800000 ]
```

I isolated the assertions so the timer evidence does not lean on the unrelated `setVoiceNetworkActive` expectation. With the spy *and* network asserts commented, the behavioural assertion fails on base on its own: `expected false to be true` after `advanceTimersByTimeAsync(31 * 60_000)`. With only the spy commented, the network assert fails (0 calls) — noted and excluded as non-timer evidence. Both timer assertions therefore fail independently on pre-fix code. `stopVadStandby()` → `expect(vadStandbyOwnsMicrophone()).toBe(false)` is the control that proves the assertion can observe the failing state.

### Blocker 2 (terminal close stranded by client pause) — FIXED

Every `close()` callsite is now forced: `:97` overflow, `:209` `finishTurn`, `:288` provider error, `:303` client stop, `:360` reconnect exhaustion, `:395` 401/403 refusal, `:404` terminal ws close, `:430` `main` catch. No unforced `close()` remains.

The part I like: `flushOutput` gates on `outputPausedByClient ? outputQueue.findIndex((item) => item.force) : 0`, so forced frames are found *anywhere* in the queue while paused — that also closes my MINOR about forced frames only bypassing at the head — **but** `index = 0` when not paused, so the normal path stays strict FIFO and `realtime.closed` still waits behind queued audio so a farewell finishes playing. Queue-jumping happens only under client pause, and only for terminal/clear frames, where dropping the tail is the correct recovery. `findIndex` returning `-1` returns without scheduling, and both `resume_output` and a subsequent forced `emit` re-arm the timer, so there is no new deadlock.

### Blocker 3 (unbounded low-latency Wi-Fi lock) — FIXED

`applyNetworkLock` (`VoiceOverlayService.kt:545-552`) releases only the Wi-Fi lock; `PARTIAL_WAKE_LOCK` and the microphone FGS are retained for background detection (`:535-543`). Transitions audited and correct: cold start → standby RELEASED; wake → realtime HELD (`vadStandby.ts:139-145` `acquireRealtimeCapture`, `realtimeSessionState.ts:359-369` before opening either the plugin or WebRTC transport); stop with standby on → RELEASED via the `active` fast path (`vadStandby.ts:52-54`); stop with standby off and `onDestroy` → both released (`:453-464,509-520`); provider failure → `rearmVadStandby` RELEASED (`:173-178`).

The ordering race is correctly fenced: `startForegroundService` is async, so JS can call `setNetworkActive(false)` before the service exists; the `@Volatile` companion flag (`:41-45`) persists the intent and `acquireVoiceLocks` re-applies it, so the lock cannot end up wrongly held. iOS is a no-op stub consistent with the neighbouring shims. Matches your `dumpsys` observation.

### Also fixed

Deadline pacing (`outputDeadline`) replaces the fixed per-frame `setTimeout`. Measured, same box as before: **0.6 % → 0.01-0.13 %** drift (50 ms → 1-10 ms over 8 s of audio), with zero zero-delay catch-up frames. Reset on empty queue, `drain`, `pause_output`, stop, overflow, and `clearQueuedAudio`-to-no-audio, so a new utterance re-anchors instead of burst-writing. The burst fixture is now ordered and non-constant (`Buffer.alloc(960, index)` + positional first-byte comparison + full PCM equality), so it can actually detect reordering. Tests at this SHA: 7/7, 5/5, 10/10 = **22 passed / 0 failed**; root typecheck exit 0; mobile typecheck shows only the 2 pre-existing `react-native-webrtc` TS2307 errors (absent from the hoisted tree, file not in this PR), zero errors in the 12 touched files, and `setVoiceNetworkActive` typechecks against `VoiceNative` and both platform declarations.

### New blocker — non-toggle callers silently disable wake on speech

Restoring `if (!armed) storage.getState().applyLocalSettings({ vadStandbyEnabled: false })` (`realtimeSessionState.ts:437`) fixed the ON-but-dead toggle, but it fires for **every** caller, and `configureVadStandby(true)` is not only called from the toggle:

```
KernelNotifications.tsx:118-122
React.useEffect(() => {
    if (isAuthenticated && appActive && vadStandbyEnabled && sessionCount > 0) {
        void configureVadStandby(true);
    }
}, [appActive, isAuthenticated, sessionCount, vadStandbyEnabled]);
```

`armVadStandby` returns `false` on transient, entirely normal states — `session !== null`, `starting`, `dictating` (`:399`), and unresolvable `target` (`:406-407`). Reachable sequence:

1. Wake on speech ON; a realtime session is live, so `session !== null`.
2. The app goes background→foreground, or `sessionCount` changes — routine while agents are working — so the effect re-runs.
3. `configureVadStandby(true)` → `armVadStandby()` returns `false` at `:399`.
4. `:437` persists `vadStandbyEnabled: false` to MMKV. No alert: the "Wake on speech unavailable" alert lives in `settings/voice.tsx:101-106`, not in the effect.
5. The `vadStandbyEnabled` dep flips, the effect stops running, and the setting is latched off.

Same outcome if the host is briefly unresolvable when the app foregrounds, or if the user is dictating. `vadArming !== null` returns the in-flight promise (`:400`), so a concurrent non-toggle call inherits the same `false` and resets too.

So wake on speech can now turn itself off without the user touching the toggle — a worse failure mode than the 30-minute TTL this PR set out to remove, because it triggers on incidental app lifecycle rather than predictably. It also breaks the acceptance criterion the PR body claims.

Minimal fix: keep the reset on the explicit path only. `settings/voice.tsx:104-106` already branches on `!armed` to alert — clear the preference there and drop `:437`. Alternatively distinguish "cannot arm right now, retry later" (session live / dictating / host down → keep the preference) from "cannot ever arm" (permission denied → clear + alert). Either way the opportunistic re-arm must never destroy the preference.

### Minor — footer now overstates on-device-only audio

`settings/voice.tsx:139` dropped the qualifier: "Audio stays on this device until speech is detected." became "**Audio stays on this device.** Wake on speech stays on until you disable it and uses additional battery." Battery and persistence are now named, which is what I asked for. But after the gate wins, the buffered pre-roll *and* live microphone audio do go to the provider — `vadStandby.ts` pre-roll → `acquireRealtimeCapture` `lease.pending` → `realtimeSession.ts:176-181` `captureFrame`. As written the first sentence claims audio never leaves the device, on a privacy-sensitive string. Restore "until speech is detected". Second sentence also reads better split, e.g. "…stays on until you disable it. It uses additional battery."

### Housekeeping

CI is **pending** at `af95fbe8` (suite + both Analyze jobs queued at review time) — I ran the three touched specs and both typechecks locally instead, but the full suite is unverified at this SHA. The branch is also `BEHIND` now that `origin/main` moved to `85b22943`; rebase before merge. No physical audible-playback claim made, and I did not re-run it — the deadline pacer plus the native 250 ms prebuffer make underrun much less likely than before, and `underruns`/`drainRestarts` remain instrumented, so a device pass is still the right final check rather than a blocker.

Fix the preference reset and the footer qualifier and I'll approve; the pacing, forced-close, and Wi-Fi-lock work is done and verified.

umeranjum17 (commented):
## APPROVE — reviewed at `845e0a7b593a5de74318df6dbaaf89a122c7cbf6` (base `8cf722275ad75e524cfcbc44b041357996c80281`)

All four blockers raised across `3746a314` and `af95fbe8` are fixed. I verified each by falsification against the pre-fix code rather than by reading the diff, so these are observations, not readings.

### Final blocker (silent preference disable) — FIXED and genuinely defended

The tri-state is the right shape. `armVadStandby(): 'armed' | 'retry' | 'failed'` maps every transient condition to `'retry'` — pref-off/dictating/`session !== null`/`starting` at `:397`, missing target and the same set re-checked inside the task at `:404-406`, epoch race at `:414-416` — and reaches `'failed'` only from a real `startVadStandby()` rejection (`:418`). `configureVadStandby` reverts solely on `'failed'` (`:437`) and returns `result !== 'failed'`, and `retryVadStandby` (`:444-447`) contains no `applyLocalSettings` at all.

Callers check out: the only production callers of `configureVadStandby` are the settings toggle (`voice.tsx:102,104`); `KernelNotifications.tsx:120` now calls `retryVadStandby()`; and the two automatic re-arm paths call `armVadStandby`/`rearmVadStandby` directly (`:302-303` on disconnect, `:290` on start failure), so none of them can touch the preference.

I went looking for a way to still reach `'failed'` transiently, since `startVadStandby` returns false on owner supersession (`vadStandby.ts:59,61,86,98`) and `startRealtimeSession` calls `cancelVadStandbyStart()` which nulls the owner. That path is fenced: `startRealtimeSession` does `vadEpoch += 1` at `:335` *before* `cancelVadStandbyStart()` at `:337`, so the in-flight arm hits the epoch check and returns `'retry'`. `'failed'` is therefore confined to stable native failures — service start, audio routing, recorder init/start — exactly as claimed.

Falsification of the new flow test (`dictation.spec.ts:220`, `it('keeps live-session standby enabled and arms it once realtime ends')`) against the `af95fbe8` state module, HEAD spec kept. All three assertions I predicted fail:

```
dictation.spec.ts:228  await expect(configureVadStandby(true)).resolves.toBe(true)
  AssertionError: expected false to be true
dictation.spec.ts:231  expect(mocks.vadStandbyEnabled).toBe(true)
  AssertionError: expected false to be true          // preference silently cleared
dictation.spec.ts:232  expect(mocks.applyLocalSettings).not.toHaveBeenCalledWith({ vadStandbyEnabled: false })
  Received: 2nd spy call: [ { vadStandbyEnabled: false } ]
```

The `retryVadStandby is not a function` error was correctly identified as a harness artifact (the export does not exist at `af95fbe8`) and excluded; the behavioural evidence was reproduced by substituting `configureVadStandby(true)`, which is literally what `KernelNotifications` called at that SHA. Making the store mock's `vadStandbyEnabled` mutable instead of a frozen `false` is what gives these assertions teeth — good call, and one flow test rather than a matrix is the right size for this repo.

### Earlier blockers — all confirmed fixed

**>30 min guard** (`af95fbe8`): fails on base at `realtimeSession.spec.ts:407` with `[Function expireVadStandby], 1800000`; the behavioural assertion also fails independently (`expected false to be true` after `advanceTimersByTimeAsync(31 * 60_000)`) once the spy and network asserts are commented, so the timer evidence stands on its own.

**Gemini terminal closes**: unchanged since `af95fbe8` and still correct — all eight `close()` sites forced (`:97,209,288,303,360,395,404,430`), with `outputPausedByClient ? findIndex(force) : 0` so forced frames jump the queue *only* under client pause while the normal path stays strict FIFO and a farewell still finishes playing.

**Low-latency Wi-Fi lock**: `applyNetworkLock` drops only the Wi-Fi lock, `PARTIAL_WAKE_LOCK` and the microphone FGS stay up for background detection, and the async-`startForegroundService` ordering race is fenced by the `@Volatile` companion flag re-applied in `acquireVoiceLocks`. Matches your `dumpsys`.

**Pacing**: deadline-based, measured 0.6 % → **0.01-0.13 %** drift (50 ms → 1-10 ms per 8 s of audio) with zero catch-up frames.

### Copy — FIXED

"Listens only on this device until speech is detected. The setting stays enabled until you disable it and uses additional battery." Accurate about the gate, and persistence and battery are both named. Restoring the "until speech is detected" qualifier matters because pre-roll and live audio genuinely do leave the device after the gate wins.

### Verification at this SHA

Focused specs 19/19 (`dictation` 7, `realtimeSession` 7, `sessionSync.integration` 5), provider 10/10, root `yarn typecheck` exit 0, mobile typecheck showing only the 2 known pre-existing `react-native-webrtc` TS2307 errors (package absent from the hoisted tree; file not in this PR), zero errors across the 14 touched files. Contract dist had to be rebuilt in the worktree first — the hoisted `@muxr/contract` resolves to a stale herdr worktree, the usual trap, not a PR defect.

### Non-blocking, for later

- Preference and armed state can still diverge in one narrow case: if a native arm failure occurs outside the toggle path — permission revoked in system settings while the preference is true — `retryVadStandby` correctly refuses to clear it, so the switch reads ON while standby is dead until the user toggles. That is the correct trade given "never auto-disable"; the clean long-term answer is to render armed state separately from intent rather than mutating the preference.
- The Gemini pacer itself still has no direct coverage (only the pure `chunkAudio` chunker, which is now ordering-sensitive). Fine by this repo's few-big-tests rule; worth one flow test through the plugin's stdout if that code churns again.
- Physical audible playback was not re-run and you did not claim it. The deadline pacer plus the native 250 ms prebuffer make underrun materially less likely than before, and `underruns`/`drainRestarts` stay instrumented, so a device pass is the right final check rather than a gate.

### Before merge

CI is **pending** at `845e0a7b` (suite and both Analyze jobs queued at review time) — my sign-off rests on the focused specs and typechecks I ran locally, so let the suite go green. The branch is also `BEHIND` now that `origin/main` moved to `85b22943`; rebase first.

Approving on that basis. The pacing, backpressure, forced-teardown, Wi-Fi-lock, and standby-persistence work is complete and each fix is backed by a test that fails without it.

umeranjum17 (commented):
## APPROVE (re-affirmed after rebase) — `fc10f9a5b7fb82a90a29a73b7157014d29fb1e14` onto `7f778b450e326381aed1372273e6e3b31c6edb1b`

Rebase-only review, as asked. The rebase is content-preserving and introduces no behavioural or integration change versus the SHA I approved (`845e0a7b`). Not merged.

### The patch is byte-for-byte the approved one

I compared the applied patches — `8cf72227...845e0a7b` versus `7f778b45...fc10f9a5` — and every `+`/`-` content line is identical. The only two deltas in the entire 787-line patch are positional metadata: a blob hash, and one hunk anchor moving `@@ -525` → `@@ -564`.

Per-file against the approved SHA, **13 of 14 files are byte-identical**, including all the load-bearing ones: `plugins/voice-gemini/stream.mjs`, `realtimeSessionState.ts`, `vadStandby.ts`, `settings/voice.tsx`, `KernelNotifications.tsx`, all four voice-overlay native files, and every spec but one.

The 14th, `sessionSync.integration.spec.ts`, is the only file main also touched. Its approved→rebased delta is *exactly* main's own edit to it — I diffed `845e0a7b→fc10f9a5` against `8cf72227→7f778b45` for that path and the content lines match, so the file absorbed main's agent-presentation rewrite without perturbing a single PR line. The four vadStandby lines still sit correctly at `:567,573,574,575`, inside the restart-simulation test and correctly ordered around `vi.resetModules()` at `:570`; main's `agentLabels`/`herdPanes` work lives in a different test at `:323-371`.

### Hygiene

Merge-base is now exactly `origin/main`, so the earlier `BEHIND` state is resolved. Exactly one commit on top (`fc10f9a5`, single parent `7f778b45`) — no merge commit, no duplicated voice commit. Zero conflict markers across all 1753 tracked files. Same 14 changed files as the approved SHA.

### No semantic conflict with #161/#162

This was the real risk, since a clean textual rebase can still break behaviour. It does not here:

- Main's contract change (`control-plane/domain/requests.ts`) is **comment-only** on `tab.close`/`pane.close`/`workspace.close`/`session.stop`. No type or signature change, and no PR file references those keys.
- `paneDisplayName`/`paneTaskTitle` — which main removed from the spec's imports — are absent from every PR production file. So are `agentLabels`/`agentPresentation`/`herdPanes`.
- `KernelNotifications.tsx` was my highest-risk file, since this PR edits it *and* it sits in the `herd/presentation` folder main reworked. It is clean: byte-identical to the approved SHA, and versus current main it contains only the intended two-line swap (`configureVadStandby(true)` → `retryVadStandby()`, and the matching import). Its other imports — `completionNotificationState`, `completionTransition`, `herdNotificationState`, `sortHerd`, `HerdNotificationState` — still resolve against main's rewritten `../domain/herd` with the same fields it reads, and `HerdPane.name` still exists. No stale references.

### Runtime at this SHA

Focused **19/19** (`dictation` 7, `realtimeSession` 7, `sessionSync.integration` 5) and provider **10/10**, corroborating your numbers. The overlap file passes 5/5 with both regions green — main's presentation assertions and the PR's persistence assertions. Root `yarn typecheck` exit 0; mobile typecheck shows only the two known pre-existing `react-native-webrtc` TS2307 errors (package absent from the hoisted tree, file not in this PR), nothing new. Contract dist had to be rebuilt in the worktree first, which matters more than usual here because main just changed contract sources — a stale hoisted dist would have produced bogus failures.

I did not re-run the full `yarn run check` (you scoped this to the rebase delta and reported 34/34 in 118.7s); my sign-off rests on the focused specs, the provider specs, both typechecks, and the patch-equality proof above.

### Before merge

CI is **pending** at `fc10f9a5` and GitHub currently reports `BLOCKED` — that is procedural, not a code problem: the branch needs an approving review, and GitHub refuses to let the author's own account approve, so my APPROVE lands as a comment. Let the suite go green, then have someone else click approve or merge under whatever override the repo allows.

Everything I approved at `845e0a7b` still holds, and the earlier caveats stand unchanged: physical audible playback remains the one thing untested by CI, and the Gemini pacer still has no direct coverage beyond the ordering-sensitive `chunkAudio` test. Neither gates this.

PR #164, opened 2026-08-29
Persist mobile lifecycle notification levels

umeranjum17:
## What

- add persisted Off, Important, and All activity lifecycle notification levels
- apply one shared admission policy to local Android presentation and relay-owned iOS push delivery
- add an accessible Settings control while keeping While You Were Away unchanged

## Why

Routine completion should not interrupt fresh installs. Blocked and failed agents remain important by default, while users can explicitly disable alerts or opt into all activity.

## Verify

```bash
yarn typecheck
yarn workspace @muxr/mobile typecheck
yarn vitest run apps/mobile/sources/catalog/application/sessionSync.integration.spec.ts apps/mobile/sources/herd/domain/herd.spec.ts --config apps/mobile/vitest.config.ts
yarn vitest run apps/relay/src/platform/persist.integration.test.ts
yarn check
```

All commands passed. Built an x86_64 release APK, installed it on an API 36 emulator, paired over `adb reverse tcp:8787 tcp:8787`, and verified Off, Important, and All activity controls plus Off persistence across restart. The flow test covers every level/event combination, persistence/restart, and unchanged recent activity.


umeranjum17 (commented):
## REQUEST CHANGES

Reviewed at exact head `b0f82982ab6185d0532f7945edf244365123ffc1` vs `origin/main` (merge-base `85b22943`), in a clean detached worktree with `packages/contract` rebuilt first. Read-only; nothing merged or edited.

One blocking defect (**F1**, native mode passthrough — acknowledged, fix in flight): the shared admission policy is applied to Android's *alerting* flag rather than to the notification itself, so `Off` does not suppress Android lifecycle notifications and the `Important` default does not suppress completion notifications. Recorded in full below with exact native evidence so the fix has a spec to land against.

Two majors follow it: a persisted-ledger side effect that makes level upgrades irreversible (**F2**) and an unfenced subscription update race (**F3**). The admission boundary itself, the relay stored schema, per-device levels, API validation and While You Were Away all check out.

---

### F1 — BLOCKING · Android lifecycle notifications are silenced, not suppressed

**`apps/mobile/sources/herd/presentation/KernelNotifications.tsx:37-48`** rewrites only `eventKey`, leaving `mode` intact:

```ts
function applyLifecycleNotificationLevel(notification, level) {
    if (notification.mode === 'attention' && !lifecycleNotificationAllowed(level, 'blocked')) {
        return { ...notification, eventKey: 'attention:' };
    }
    if (notification.mode === 'finished' && !lifecycleNotificationAllowed(level, 'done')) {
        return { ...notification, eventKey: '' };
    }
    return notification;
}
```

`mode` — not `eventKey` — is what decides whether a notification is posted. Exact chain:

**1. JS forwards both verbatim.** `apps/mobile/modules/voice-overlay/index.ts:146-159`

```ts
export function updateVoiceNotification(
    herd: { mode: HerdMode; count: number; name: string; names: string; eventKey: string },
    ...
    return native?.updateNotification(
        herd.mode,        // :153
        herd.count,
        herd.names,
        herd.eventKey,    // :156
```

**2. Native consumes `eventKey` only to compute an alert flag.** `apps/mobile/modules/voice-overlay/android/src/main/java/expo/modules/voiceoverlay/VoiceOverlayService.kt:103-128`

```kotlin
herdMode = mode                                   // :103
herdEventKey = eventKey.trim().take(200)          // :107
val newEventAlert = when (mode) {
  "attention" -> {
    val current = herdEventKey.removePrefix("attention:").split(",").filter(String::isNotBlank).toSet()   // :110
    val alert = current.any { it !in lastAttentionKeys }                                                  // :111
    ...
  }
  "finished" -> {
    val alert = herdEventKey.isNotBlank() && herdEventKey != lastFinishedKey                              // :118
    ...
  }
}
pendingEventAlert = if (mode == "attention" || mode == "finished") pendingEventAlert || newEventAlert else false  // :128
```

**3. That flag only controls sound and heads-up.** `VoiceOverlayService.kt:335,346-347`

```kotlin
val event = !activeVoice && pendingEventAlert   // :335
    .setOnlyAlertOnce(!event)                   // :346
    .setSilent(!event)                          // :347
```

**4. Posting is decided by `herdMode`.** `VoiceOverlayService.kt:588-606`

```kotlin
} else if (herdMode == "working" || herdMode == "attention") {
    manager(this).notify(HERD_NOTIFICATION_ID, buildNotification(this, false))   // :593
    ...
    stopForeground(STOP_FOREGROUND_REMOVE)
    if (herdMode == "finished") {
        manager(this).notify(HERD_NOTIFICATION_ID, buildNotification(this, false))   // :602
    } else {
        manager(this).cancel(HERD_NOTIFICATION_ID)                                    // :605
    }
```

The `cancel` branch at `:605` is the only suppression path, and no level ever routes to it. Same shape in the service-absent path at `:177-181`.

Traced the two functions verbatim at every level — `mode` survives unchanged:

```
off       finished -> {"mode":"finished","count":1,"names":"Maria","eventKey":""}
off       attention-> {"mode":"attention","count":1,"names":"Maria","eventKey":"attention:"}
important finished -> {"mode":"finished","count":1,"names":"Maria","eventKey":""}
all       finished -> {"mode":"finished","count":1,"names":"Maria","eventKey":"finished:maria"}
```

**Observable result on Android**, with copy from `VoiceOverlayService.kt:288-300`:

- Level `Off`, blocked agent → **"Maria needs you / Open muxr to respond"** still posted, with the `Open` / `Talk` actions (`:361-364`). Silent, but present.
- Level `Off` **and** the `Important` default, completed agent → **"Maria finished / Work completed"** still posted. Reachable whenever `lifecycleCatalogAvailable === false`, i.e. hosts that do not publish the lifecycle catalog — `KernelNotifications.tsx:83-94` selects `completionNotificationState` on exactly that branch.

It is worse at the moment of choosing. `visibleSignature()` (`VoiceOverlayService.kt:142`) includes `herdEventKey`, so flipping the level changes the signature and *actively reposts* the same card via `doRequestPost` → `refreshNotification`. Selecting `Off` re-fires the notification the user just turned off.

Violates "Off suppresses all lifecycle push/local notifications" and "Important admits blocked+failed", and defeats the PR's own stated goal that "routine completion should not interrupt fresh installs".

**Scope:** Android only. `voice-overlay/index.ts:50-51` (`Platform.OS === 'web' ? null : requireOptionalNativeModule('VoiceOverlay')`) resolves no module on iOS or web — only `android/` sources exist — so `updateVoiceNotification` no-ops there, and the iOS/local-Expo path through `pendingLifecycleEvents` is correctly gated.

**Fix direction:** map the **mode**, not the key. `finished` + not allowed → a mode that reaches `manager.cancel(HERD_NOTIFICATION_ID)` (e.g. `idle`). `attention` + not allowed → `working`, which keeps `promote` true (`:336`) so the mic foreground service stays legal while dropping the "needs you" copy and actions. Please also confirm the level-change repost no longer surfaces a card.

---

### F2 — MAJOR · Suppression writes the persisted dedup ledger, so raising the level can never release history

**`apps/mobile/sources/watch/application/agentWatch.ts:285-307`**

```ts
presentedIds.add(event.eventId);                    // :299
suppressed.push({ eventId: event.eventId, at: event.at });
...
notification.presented = boundPresented([...notification.presented, ...suppressed]);   // :304
saveNotifications();                                                                    // :305
```

`applyCatalog` runs this across the whole bounded `lifecycleEvents` list on every reconciliation (`:399-400`), and `applyEvent` per event (`:418-419`). Consequences:

- `Off` → `Important`/`All` never surfaces blocked/failed that arrived while Off — including an agent that is *still blocked right now*.
- `Important` → `All` never surfaces already-suppressed `done`.
- What was a pure predicate (`lifecycleNeedsNotification`) is now a persisting MMKV write on a reconnect-triggered path.

Eviction itself is safe: `boundPresented` (`:115-125`) and `boundLifecycleEvents` (`:102-113`) share `MAX_EVENTS = 50` with the same recency ordering, so suppressed records cannot evict a dedup record that is still reachable from the catalog. The problem is semantic, not capacity.

This is plausibly intended ("do not retroactively alert"), but nothing states it and no test pins it, so it reads as a bug on next contact. Please comment the intent on `admitNotifications` and add one assertion that a level upgrade does not replay suppressed history. If it is *not* intended, suppression must not touch `presented` at all — filter without recording.

---

### F3 — MAJOR · Unfenced subscription update: relay level can disagree with the persisted local setting

**`apps/mobile/sources/app/(app)/settings/notifications.tsx:22-26`**

```ts
const select = (next: LifecycleNotificationLevel) => {
    if (next === level) return;
    setLevel(next);
    void updateNativePushNotificationLevel(next);   // :25 — result discarded
};
```

**`apps/mobile/sources/utils/nativePushNotifications.ts:27-54`** has no generation counter and no serialization, and `PushService.subscribeExpo` (`apps/relay/src/push/infrastructure/push.ts:155-165`) replaces the token record on arrival order. Two quick taps (`Off` then `All`) can land reversed, leaving the relay storing `off` while the device persisted `all`.

`registerNativePushNotifications` has a `registering` single-flight guard (`nativePushNotifications.ts:59`) that `updateNativePushNotificationLevel` does not participate in, and the register path runs on every foreground (`KernelNotifications.tsx:135-139`) — so a foreground re-register can interleave with an in-flight level update as well.

Partial mitigation already present: that foreground re-register unconditionally re-POSTs the current level (`nativePushNotifications.ts:80-82`), so divergence self-heals on the next foreground. Until then iOS push filtering silently disagrees with the UI, and a failed POST produces no user-visible signal at all.

Fix direction: route level updates through the existing `registering` chain, or a monotonic request id where only the newest response is honoured; surface persistent failure rather than swallowing it.

---

### F4 — Minor · Level is only evaluated at ingest; a downgrade does not drain the queue

`apps/mobile/sources/catalog/application/sync.ts:354-381` iterates `pendingLifecycleEvents` with no level check, so switching to `Off` while a `done` is already admitted still posts it. Window is small — presentation is kicked immediately after `applyLifecycleEvent`/`applyLifecycleCatalog` (`:313`, `:501`) and `pendingLifecycleEvents` is not persisted across restart — but it is a real hole in "Off suppresses all".

---

### F5 — Minor · Duplicated accessibility state

`apps/mobile/sources/app/(app)/settings/notifications.tsx:46` bakes `Selected` / `Not selected` into `accessibilityLabel`, while `apps/mobile/sources/components/Item.tsx:275` already emits `accessibilityState={{ disabled, selected }}`. The active row announces its state twice. Drop it from the label.

Out of scope here, noted for completeness: the group is semantically a radio set but `Item` hardcodes `accessibilityRole="button"` (`Item.tsx:274`).

---

### `applyAttentionCatalog` — classification: latent bypass, not a blocker at this head

`apps/mobile/sources/catalog/application/sync.ts:339-352` is a second local-notification emitter that never consults `lifecycleNotificationLevel`. Traced every platform against both guards:

```ts
if (previous.has(entry.sessionId) || Platform.OS === 'ios'
    || (Platform.OS === 'android' && entry.reason === 'done')) continue;   // :345
...
void this.scheduleSessionNotification(entry.sessionId, `${name} needs attention.`);
```
```ts
private async scheduleSessionNotification(sessionId: string, body: string) {
    if (Platform.OS === 'android') return;                                  // :388
```

- **iOS** — skipped by the `Platform.OS === 'ios'` guard at `:345`. Unreachable.
- **Android, `reason === 'done'`** — skipped by the same guard. Unreachable.
- **Android, other reasons** — reaches `scheduleSessionNotification`, returns immediately at `:388`. Unreachable.
- **Web** — both guards pass, so `Notifications.scheduleNotificationAsync` is called at `:392`, un-gated by level. Requires `lifecycleCatalogAvailable === false` and `AppState.currentState !== 'active'`. Web has no level control at all: `SettingsView.tsx:457` hides the Agent notifications row on web, and `Platform.OS === 'web'` keeps the separate Web Push toggle at `:499-508`.

**Verdict: not a blocker — dead on both mobile platforms at this head.** It is a latent bypass: the level boundary holds here by platform accident, not by construction, so relaxing either guard silently loses enforcement. Recommend routing it through `lifecycleNotificationAllowed` keyed on `entry.reason` so enforcement is structural. Note `AttentionReason` is `['waiting', 'blocked', 'failed', 'done']` (`packages/contract/src/herd/domain/sessionState.ts:327`) and `'waiting'` has no `AgentLifecycle` counterpart — that mapping needs an explicit decision rather than a cast.

---

### Migration findings

- **Absent setting → Important: satisfied.** `apps/mobile/sources/catalog/application/localSettings.ts:19` adds `lifecycleNotificationLevel: z.enum(LIFECYCLE_NOTIFICATION_LEVELS)`, `:50` defaults it to `'important'`, and `localSettingsParse` (`:59-65`) spreads `localSettingsDefaults` under the partial parse so a stored blob without the key resolves to `important`. Pinned by `sessionSync.integration.spec.ts:395`.
- **Relay-side absent level → Important: satisfied.** `level?` is optional on `ExpoPushTokenRecord` (`apps/relay/src/push/infrastructure/push.ts:74`), read as `entry.level ?? 'important'` at the delivery filter (`:209`), and `load()` (`:134-140`) tolerates records written before this PR. `subscribeExpo` also defaults the parameter to `'important'` (`:158`). Verified by the deliberately unchanged `subscribeExpo(account.accountId, 'ExpoPushToken[fixture-token]')` call at `persist.integration.test.ts:206` still passing.
- **Explicit legacy off → Off: no source exists to migrate from.** `git grep -E "(notifications?|alerts?)(Enabled|Off|Muted|Disabled)"` across `apps` and `packages` returns only OS-permission and Android Live-Updates *UI* state (`SettingsView.tsx:160,163,250-255`), none of it persisted; there is no legacy notification boolean in `LocalSettingsSchema`, in `persistence.ts`, or in the relay's stored schema. Either that requirement is stale or its source of truth lives outside this repo. Please state which in the PR body — as it stands the requirement is silently unaddressed, and a reviewer cannot distinguish "nothing to migrate" from "migration forgotten".
- **Forward-compat note:** `localSettingsParse` returns `{ ...localSettingsDefaults }` wholesale when `safeParse` fails (`:61-63`), so a value outside the enum — e.g. written by a newer build and then downgraded — resets *every* local setting, including `themePreference` and `savedLayouts`. Pre-existing hazard (`themePreference` has the same shape), but this PR adds a second enum that a future release is likely to extend. Consider per-field fallback.

---

### Subscription findings

- **Per-device levels work on both routes.** Account-token route passes no `deviceId` (`apps/relay/src/httpHandlers.ts:395`), and the replace-filter at `push.ts:161` — `entry.token !== token && (deviceId === undefined || entry.deviceId !== deviceId)` — preserves every other token, so two iPhones on one account hold independent levels keyed by push token. Local-pairing route keys by `device.deviceId` (`apps/relay/src/relay.ts:899`). `removeExpoToken` / `removeExpoDevice` stay matched to their respective routes.
- **Validation rejects before any write.** `parseLifecycleNotificationLevel` (`packages/contract/src/herd/domain/sessionState.ts:175-179`) returns `undefined` for anything not in the enum, and both routes 400 before touching state (`httpHandlers.ts:387-393`, `relay.ts:897-899`). Token validation still precedes the method branch; `DELETE` correctly ignores `level`.
- **Client URL bases match.** `subscribeNativePush` uses `relayControlUrl(getCachedConnectionSettings().relayUrl)` (`nativePushNotifications.ts:35`), identical to the base `registerNativePushNotifications` (`:68`) and `unregisterNativePushNotifications` (`:96`) already compute. The refactor did not shift the endpoint.
- **`Off` keeps the device subscribed with `level: 'off'`** rather than unsubscribing, and relies on the relay filter (`push.ts:208-209`). Correct choice — re-enabling needs no new OS permission round-trip — and consistent with `updateNativePushNotificationLevel` deliberately never prompting (`:44-54`).
- **Pre-existing bug fixed en route.** `push.ts:239-242` now maps `DeviceNotRegistered` tickets back to `eligibleExpo` tokens instead of indexing the unfiltered `expo` list. Without this the new filter would have pruned the wrong devices on every partial-suppression batch. Good catch, and load-bearing for the feature.
- **Web Push is not level-filtered.** `push.ts:198-205` sends `this.subs` unconditionally; only `expoSubs` is filtered. Defensible — web has its own on/off control and the Agent notifications row is hidden there — but it means `Off` is Expo-scoped, not account-scoped. Worth one sentence in the PR body.
- **Fully suppressed events are not recorded as delivered.** `push.ts:247-256` only pushes to `deliveredEvents` when `sent > 0`. No retry storm: the host is fire-and-forget and ignores the outcome (`apps/host/src/agent/infrastructure/herdrSessionSource.ts:586-594`). But a re-notify of the same `eventId` after a level upgrade would deliver late. Fine as a decision, not as an accident.

---

### Verified correct

- **One admission boundary, correct at all nine (level × terminal state) combinations.** `packages/contract/src/herd/domain/sessionState.ts:182-185`: `off` false everywhere, `important` = blocked|failed, `all` adds done, non-terminal states never admitted. Genuinely shared by device-local admission, the Android presentation transform and relay push delivery — three callsites, one function. `admitNotifications` correctly reuses `lifecycleNotificationAllowed('all', ...)` as the "notifiable at all" predicate, exactly replacing the deleted `lifecycleNeedsNotification`.
- **Clean cutover.** `lifecycleNeedsNotification` removed from `apps/mobile/sources/watch/domain/lifecycle.ts` and its re-export in `watch/index.ts`; repo-wide grep finds zero residual references. All five `subscribeExpo` callsites migrated. Both barrel exports updated (`packages/contract/src/herd/index.ts`, `packages/contract/src/index.ts`).
- **While You Were Away is genuinely untouched.** `unseenActivityRows` reads `state.lifecycleEvents` with a seen-set from a separate AsyncStorage key `muxr.herd.seen-activity.v1` (`apps/mobile/sources/herd/application/useActivityAcknowledgements.ts:4`), with no relationship to `notification.presented`. `admitNotifications` never mutates `lifecycleEvents`. The new test asserts all nine events stay visible at every level (`sessionSync.integration.spec.ts:422-429`).
- **No permission or provider regression.** `requestNotificationPermission(false)` still gates `updateVoiceNotification` (`KernelNotifications.tsx:111`); `updateNativePushNotificationLevel` deliberately prompts nothing, matching its docstring; `handleIosNotifications` unchanged; the iOS row retitle at `SettingsView.tsx:492-493` is copy-only. No new fields in the push payload, so no new PII surface.

---

### Tests

Right file and right granularity — one flow test added to the existing integration spec, consistent with `AGENTS.md`. Two reservations:

1. The body is a 3×3 permutation matrix (`sessionSync.integration.spec.ts:401-420`) — the "test per permutation" shape the repo policy rejects — and it exercises the pure boundary through the store rather than a user-visible outcome. The assertions that earn their place are the persistence round-trip (`:395-398`) and the activity-unchanged check (`:422-429`); the matrix could collapse to those plus a single `important` case.
2. Nothing covers F1, and there is no clean seam for the Android presentation path without a native fake. Please state that gap explicitly in the PR body so it is a known one rather than an implied claim of coverage.

---

### Verification I ran

Clean worktree at `b0f8298`, `packages/contract` rebuilt first — a stale `dist` otherwise masks the new exports and turns every failure into a false negative:

- `npx tsc --build` → clean
- `apps/mobile` `tsc --noEmit` → only pre-existing `react-native-webrtc` module-resolution errors; that package is absent from the install entirely, unrelated to this PR
- `vitest run apps/mobile/sources/catalog/application/sessionSync.integration.spec.ts` → 6/6
- `vitest run apps/relay/src/platform/persist.integration.test.ts` → 5/5
- `vitest run apps/mobile/sources/herd/domain/herd.spec.ts apps/mobile/sources/architecture.spec.ts` → 7/7

Skipped formatter and linter as requested.

---

### Mechanics

`gh` reports `mergeable: CONFLICTING` against `main` — needs a rebase before it can land.

Filed as a comment because self-review cannot carry a formal verdict; treat the header as the verdict. F1 is the only merge blocker; F2 and F3 should land with it. Will re-review the rebased head.

umeranjum17 (commented):
## REQUEST CHANGES

Re-reviewed at `ea0b5608a89c0212e3058f530942238ed88371ac`, merge-base `bdc69fea` = `origin/main`, `mergeable: MERGEABLE`. Delta-only pass as agreed. Read-only; not merged.

**Delta scope is exactly two files** — `apps/mobile/sources/utils/nativePushNotifications.ts` (+48/−23) and its integration spec (+69/−40). Confirmed nothing else moved:

```
git diff 8cfb3665 ea0b5608 -- apps/mobile/sources/herd  → 0 lines
git diff 8cfb3665 ea0b5608 -- apps/relay packages/contract → 0 lines
```

The logout-race work is correct and well-proven. **One blocker remains, untouched by this delta: B1.**

---

### B1 — STILL BLOCKING (unchanged since `8cfb3665`) · Level `Off` stops the Android dataSync foreground service

`apps/mobile/sources/herd/domain/herd.ts:46-54` at `ea0b5608` is byte-identical to the previous head:

```ts
const filtered = (level === 'off'
    && (notification.mode === 'working' || notification.mode === 'attention' || notification.mode === 'finished'))
    || (notification.mode === 'finished' && !lifecycleNotificationAllowed(level, 'done'));
return filtered ? IDLE_NATIVE_NOTIFICATION : notification;
```

`working` and `attention` are the two FGS-eligible modes, so mapping them to `idle` routes native into the teardown branch:

- `VoiceOverlayService.kt:139` — `if (mode != "working" && mode != "attention") herdKeepalive = false`
- `VoiceOverlayService.kt:600-620`, `voiceState == "disconnected"`, `herdMode == "idle"` → else branch: `herdKeepalive = false` (`:611`), `stopForeground(STOP_FOREGROUND_REMOVE)` (`:612`), `cancel(HERD_NOTIFICATION_ID)` (`:617`), `stopSelf()` (`:619`)
- That is the `FOREGROUND_SERVICE_TYPE_DATA_SYNC` service from `:583-587`, whose own comment at `:419-421` states its purpose: *"become foreground as dataSync so Android never freezes the process and the herd socket stays live"*

JS cannot recover it: `KernelNotifications.tsx:103` guards on `!keepalive.current`; the latch is already `true` and `:93` never resets it because `herdActive` (`:65`) reads the **unfiltered** herd. `voice-overlay/index.ts:96-97` documents that a background restart would be rejected anyway (Android 12+). Recovery only on next app foreground (`:114`).

Net effect unchanged: agents working or blocked, no live voice session, user selects **Off** → the service stops, Android freezes the process, the herd socket dies in background until the app is reopened. A notification preference must not double as a background-disconnect switch.

Still pinned as intended by `sessionSync.integration.spec.ts`:

```ts
expect(nativeLifecycleNotificationState(workingNativeState, 'off')).toEqual(idleNativeState);
```

Fix as previously specified: drop the `working` clause entirely (not an alert — `herdTitle():294` "is working", `herdBody():303` "Work in progress", `pendingEventAlert` forced false at `:132`, single "Talk" action at `:370`); map `attention` → `working` rather than `idle` so `promote` (`:340`) and the FGS branch (`:605`) survive while the "needs you" title (`:293`), body (`:302`) and Open/Talk pair (`:365-368`) are dropped; keep `finished` → `idle`, which is correct because pre-PR `finished` already hit the same `:607-619` teardown. Then replace that assertion with the FGS-preserving invariant.

Neither `yarn check` nor the API36 build can catch this — the teardown lives in the native mode branch and the only test covering it asserts today's behaviour as correct.

---

### No other blockers or majors.

---

### Delta verified correct

**M1 — unregister fence.** Traced every interleaving; the ordering claim holds.

- `unregisterNativePushNotifications:131-158` runs its fence **synchronously** before any await: generation bump (`:134`) and `pendingNotificationLevel = null` (`:135`). That is what makes the fence effective rather than advisory.
- `:133` makes concurrent logouts idempotent.
- `:137` awaits `registering`, `:139` awaits `syncingNotificationLevel`, with defensive pending clears at `:138`/`:140`. The DELETE at `:145` is therefore the last relay mutation, and `clearRegisteredPushToken` (`:154`) runs after it.
- `updateNativePushNotificationLevel:84` returns `unregistering.then(() => false)` **without** setting `pendingNotificationLevel` — so a late update strands nothing for a subsequent drain to pick up. Correct choice.
- `registerNativePushNotifications:93` defers to `unregistering.then(() => registerNativePushNotifications())`, so no new registration can start inside the window — which is what closes the "new `registering` between `:137` and the DELETE" hole.
- `drainNotificationLevel:56` also fences on `generation === notificationRegistrationGeneration`, so an in-flight loop stops issuing after the bump. An already-awaiting POST at `:66` is not cancelled, but `:139` awaits it, so DELETE still lands last.
- Real callsite is correctly ordered: `AuthContext.tsx:55` awaits the unregister before the `/v1/session` DELETE (`:58`) and `clearLocalSession` (`:64`), so the credential is still valid when the push DELETE goes out.

One benign observation, not a defect: `registerNativePushNotifications` checks generation once at `:105` and not again before the POST at `:120`, so a logout racing a registration can still emit one subscribe before the DELETE. Ordering and final state are correct (DELETE wins); it is a wasted round-trip only.

**M3 — settlement recheck.** The stranded-update gap is closed on both sides:

- `:48-51` — joining an in-flight drain now chains `syncingNotificationLevel.then(() => pending === null ? true : drainNotificationLevel())` instead of returning the shared promise bare, so a level set during the settle-to-`finally` window starts a fresh drain.
- `:77-78` — the owning drain re-checks pending after settlement and chains another drain.
- `:73-76` — the `if (syncingNotificationLevel === syncing)` self-identity guard stops a settling drain from clearing a newer drain's slot. That is the subtle one and it is right.
- Recursion terminates: each drain iteration nulls `pendingNotificationLevel` at `:58`, and each recursion is gated on it being non-null.

**Test — single deferred flow, covers both claims.** `nativePushNotifications.integration.spec.ts:41-84` holds one POST open and asserts, in one flow:
- latest-level-wins — `mutations` is `[POST off, POST all]`, with `important` coalesced away;
- POST-in-flight → logout → `mutations` is exactly `[POST important, DELETE]`, the late `updateNativePushNotificationLevel('all')` resolves `false` and emits nothing, and `clearRegisteredPushToken` fires exactly once.

Right granularity and right file per `AGENTS.md` — one flow test, real module, no mock that would pass if the code broke.

---

### Verification I ran at this head

Fresh worktree, `packages/contract` rebuilt first:

- `npx tsc --build` → clean
- `apps/mobile` `tsc --noEmit` → only the pre-existing `react-native-webrtc` resolution errors (package absent from the install; unrelated)
- `vitest run` on `nativePushNotifications.integration.spec.ts`, `sessionSync.integration.spec.ts`, `architecture.spec.ts` → 13/13

CI at `ea0b5608`: `CodeQL`, `Analyze (actions)`, `Analyze (javascript-typescript)` all `success`; `suite` still `in_progress` at time of writing — please confirm it goes green, though your local `yarn check` result covers the same ground.

Trusting your reported `yarn check` and API36 x86_64 rebuild/install rather than re-running them.

---

Filed as a comment because self-review cannot carry a formal verdict; treat the header as the verdict. The delta is good and I have no further comments on it. B1 is the sole remaining blocker and needs roughly a five-line change in `nativeLifecycleNotificationState` plus the swapped assertion. Not merging.

umeranjum17 (commented):
## APPROVE

Reviewed `ea0b5608..82f1a6cf` at head `82f1a6cf267da84f019401fd5e3c99bdf935c153`, merge-base `bdc69fea` = `origin/main`, `mergeable: MERGEABLE`. Delta-only pass as agreed. Read-only; not merged.

**B1 is fixed. No remaining blockers or majors.** This is my last request-free pass — the feature is correct end to end across mobile admission, Android native presentation, and relay-owned push.

Delta scope is exactly the two expected files (`herd.ts` +17/−4, `sessionSync.integration.spec.ts` +10/−2). No drift into already-verified code:

```
git diff ea0b5608 82f1a6cf -- apps/relay packages/contract apps/mobile/sources/utils → 0 lines
```

---

### B1 — resolved

`apps/mobile/sources/herd/domain/herd.ts:46-62` now branches per mode instead of collapsing three modes into `idle`:

```ts
if (notification.mode === 'finished' && !lifecycleNotificationAllowed(level, 'done')) {
    return IDLE_NATIVE_NOTIFICATION;
}
if (notification.mode === 'attention' && !lifecycleNotificationAllowed(level, 'blocked')) {
    return { ...notification, mode: 'working', name: '', names: '', eventKey: `working:${notification.eventKey}` };
}
return notification;
```

Traced the mapping against the native renderer across every mode × level:

```
attention     off       -> mode=working    notify HERD (FGS kept)         silent=true  promote=true  act=Talk      "An agent is working / Work in progress"
attention     important -> mode=attention  notify HERD (FGS kept)         silent=false promote=true  act=Open+Talk "Maria needs you / Open muxr to respond"
attention x3  off       -> mode=working    notify HERD (FGS kept)         silent=true  promote=true  act=Talk      "3 agents working / Work in progress"
working       off       -> mode=working    notify HERD (FGS kept)         silent=true  promote=true  act=Talk      "Maria is working / Work in progress"
finished      important -> mode=idle       stopForeground+cancel+stopSelf silent=true  promote=false
finished      all       -> mode=finished   stopForeground+notify+stopSelf silent=false promote=false "Maria finished / Work completed"
```

Each requirement of the fix confirmed against the native source:

- **dataSync FGS preserved.** `attention` at `Off` now arrives as `working`, so `VoiceOverlayService.kt:139` leaves `herdKeepalive` set and `refreshNotification:604` takes the `manager.notify(HERD_NOTIFICATION_ID, ...)` path. The `:611-619` teardown (`herdKeepalive = false` / `stopForeground(STOP_FOREGROUND_REMOVE)` / `cancel` / `stopSelf`) is no longer reachable from a level change on an active herd. That was the whole blocker.
- **Genuinely non-alerting.** `:112-130` routes `working` to the `else ->` branch, and `:132` forces `pendingEventAlert = false` for any mode that is not `attention`/`finished`. So `buildNotification`'s `event` (`:339`) is false → `setOnlyAlertOnce(true)` / `setSilent(true)` (`:350-351`). No sound, no heads-up.
- **No lifecycle copy or identity leaked.** Blanking `name`/`names` makes `herdName.ifBlank { "An agent" }` (`:293-294`) render **"An agent is working"** rather than "Maria is working", and `herdBody:303`'s `herdNames.isNotBlank()` guard falls through to **"Work in progress"**. The "needs you" title (`:293`), "Open muxr to respond" body (`:302`) and the `Open`+`Talk` action pair (`:365-368`) are all gone — `:369-370` gives the plain `Talk` action. This is stricter than the direction I gave, which would have kept the agent name; the blanking is the better call.
- **Ongoing card stays legitimate.** `promote` (`:340`) is still true for `working`, so `setOngoing` / `setRequestPromotedOngoing` hold and the notification remains a valid FGS anchor rather than a dismissible card.
- **Ordinary `working` is untouched.** The test asserts referential identity (`.toBe(workingNativeState)`), so it is a true passthrough with no clone — the over-reach is gone, not merely reshaped.
- **`finished` unchanged and still correct.** Pre-PR `finished` already hit the same `:607-619` branch, so routing it to `idle` regresses nothing; `all` still posts the completion card.
- **`connecting` / `offline` / `idle` fall through `return notification` at every level.** Confirmed level-invariant, so "Connecting to host" (`:297`) and "Host unreachable" (`:296`) notices are untouched — as intended.
- **Event-key prefix is load-bearing, not cosmetic.** `working:${eventKey}` keeps the suppressed-attention key (`working:attention:s1`) distinct from a real working key (`working:s1`), so `visibleSignature()` (`:146`) changes on a level flip and the card actually re-renders to the neutral copy instead of being deduped against what was on screen. Plain `working:${ids}` would have collided.

### The new assertion is load-bearing

I verified the "would fail the prior idle teardown" claim rather than taking it on trust: reverting only `nativeLifecycleNotificationState` to the pre-B1 filter in a scratch tree fails the flow test with exactly the right diff.

```
× session sync flow > persists lifecycle notification levels without changing recent activity
  → expected { mode: 'idle', count: +0, …(3) } to deeply equal { mode: 'working', count: 1, …(3) }
  Tests  1 failed | 5 passed (6)
```

So the assertion defends the FGS-preserving invariant and would catch a regression here, which is exactly what was missing at `8cfb3665`.

### One deliberate distinction, noted as correct

Suppressing `attention` at `Off` resets `lastAttentionKeys` to empty (`:127`), so switching back to `Important` while an agent is *still blocked* does alert. That is right and does not contradict the F2/F4 "downgrade claims events permanently" semantic: `pendingLifecycleEvents` is an *event* ledger and stays claimed, whereas the native attention card is *present state* — "an agent needs you right now". A user who just enabled alerts with a blocked agent on screen should be told. Recording it so a future reader does not mistake it for backlog release.

---

### Verification at this head

Fresh worktree, `packages/contract` rebuilt first:

- `npx tsc --build` → clean
- `apps/mobile` `tsc --noEmit` → only the pre-existing `react-native-webrtc` resolution errors (package absent from the install; unrelated)
- `vitest run` on `sessionSync.integration.spec.ts`, `nativePushNotifications.integration.spec.ts`, `herd/domain/herd.spec.ts`, `architecture.spec.ts` → 14/14
- Mutation check above

CI at `82f1a6cf`: all checks green.

Trusting your reported `yarn check` rather than re-running it.

---

### Cumulative state

Everything raised across the four passes is now closed: the admission boundary and its single `lifecycleNotificationAllowed` source of truth; the `lifecycleNeedsNotification` cutover; persistence and absent-value defaults on both mobile and relay; backwards-compatible `ExpoPushTokenRecord`; per-device levels on both subscribe routes; request validation rejecting before writes; the `DeviceNotRegistered` index-misalignment fix; Android local presentation (B1) with the dataSync FGS intact; serialized/coalesced level writes with a fenced deferred-logout flow; downgrade pruning with permanent claiming; both `scheduleNotificationAsync` paths and the legacy `watch.settled` path structurally gated; While You Were Away untouched; no notification-permission or provider regression.

Two items remain open by choice, neither blocking, both worth one line in the PR body before merge: `Off` is Expo-scoped rather than account-scoped since Web Push is not level-filtered (`push.ts:198-205`), and `waiting` attention entries no longer notify on web (`sync.ts:359`).

Filed as a comment because self-review cannot carry a formal verdict; treat the header as the verdict. Approving on my side — no further changes requested. Not merging, per your instruction.

PR #166, opened 2026-08-29
Restore mobile agent provider logos

umeranjum17:
## What

- restore the pre-regression mobile provider PNG set and alias mappings
- retain the current themed Pi assets and corrected OMP asset
- keep initials only for unknown agent kinds
- restore asset provenance and add one alias/fallback flow test

## Why

PR #162 removed the bundled provider images and collapsed supported providers to initial tiles.

## Verify

```bash
npx vitest run apps/mobile/sources/components/agentImageKind.spec.ts --config apps/mobile/vitest.config.ts
yarn workspace @muxr/mobile typecheck
yarn run check
```

`yarn run check`: `34/34 passed in 133.3s`

Built the API 36 x86_64 release APK, installed it on the API 36 emulator, and opened the real New agent picker. The accessibility tree exposes all 21 provider buttons; the captured device screenshot shows image assets for each supported provider.

PR #167, opened 2026-08-29
Require explicit realtime prompt targets

umeranjum17:
## Summary
- require a nonempty explicit Agent Name or Task Title for realtime `prompt_agent` mutations
- validate Herdr `agent.prompt` receipts against the resolved pane and report only `Queued`
- journal privacy-safe provider, semantic target, and queued/rejected/failed outcomes
- keep Codex fail-closed when it cannot supply an explicit semantic target

## Verification
- `npx vitest run plugins/voice/providerRefusal.spec.mjs` (11/11)
- `node apps/host/dist/agent/infrastructure/layoutSelfCheck.js`
- real Herdr two-agent E2E: B observed `PROMPT_ROUTE_E2E_B_ONLY_20260829_0847`; A did not
- `yarn run check` (34/34)

### DIFF
--- a/apps/host/src/agent/infrastructure/herdrSessionSource.ts
+++ b/apps/host/src/agent/infrastructure/herdrSessionSource.ts
@@ -49,6 +49,7 @@ import { PluginStreamManager } from './pluginStreamManager.js';
 import {
     RealtimeCodingCoordinator,
     type RealtimeCodingAgent,
+    type RealtimePromptDiagnostic,
 } from './realtimeCoordinator.js';
 import type { HostedMachineKeys } from '../../machine/index.js';
 import type { PeerBroker } from '../../peer/index.js';
@@ -148,6 +149,8 @@ export interface CreateHerdrSessionSourceOptions {
     hostedE2ee?: HostedMachineKeys;
     /** Issues one revocable capability only to an approved voice.session child. */
     peerBroker?: PeerBroker;
+    /** Writes bounded semantic prompt outcomes to the owner-only host diagnostics journal. */
+    onRealtimePromptDiagnostic?: (event: RealtimePromptDiagnostic) => void;
 }
 
 /** herdr agent record (agent.list / snapshot.agents / agent.start result). */
@@ -173,6 +176,33 @@ export async function sendKeysToLiveAgent(
 ): Promise<void> {
     if (!agentsByPane.has(record.paneId)) throw new Error(`${record.agentName} is not ready for keys.`);
     await client.call('agent.send_keys', { target: record.paneId, keys });
+}
+
+export async function promptHerdrAgent(
+    client: Pick<HerdrClient, 'call'>,
+    record: Pick<AgentIdentity, 'paneId' | 'agentName'>,
+    text: string,
+): Promise<void> {
+    const receipt = await client.call<unknown>('agent.prompt', { target: record.paneId, text });
+    const result = typeof receipt === 'object' && receipt !== null && !Array.isArray(receipt)
+        ? receipt as Record<string, unknown>
+        : undefined;
+    const agent = typeof result?.agent === 'object' && result.agent !== null && !Array.isArray(result.agent)
+        ? result.agent as Record<string, unknown>
+        : undefined;
+    if (result?.type !== 'agent_prompted'
+        || typeof agent?.terminal_id !== 'string'
+        || typeof agent.agent_status !== 'string'
+        || typeof agent.workspace_id !== 'string'
+        || typeof agent.tab_id !== 'string'
+        || typeof agent.pane_id !== 'string'
+        || typeof agent.focused !== 'boolean'
+        || typeof agent.revision !== 'number'
+        || !Number.isSafeInteger(agent.revision)
+        || agent.revision < 0
+        || agent.pane_id !== record.paneId) {
+        throw new Error(`${record.agentName} prompt was not queued by Herdr.`);
+    }
 }
 
 function agentRouteError(code: 'agent-unavailable' | 'agent-route-ambiguous'): Error {
@@ -382,7 +412,7 @@ export async function createHerdrSessionSource(
             status: async (sessionId) => statusFor(sessionId).agentStatus ?? 'unknown',
             watch: waitForAgent,
             focus: focusSession,
-        });
+        }, options.onRealtimePromptDiagnostic);
         await codingCoordinator.start();
         pluginStreams = new PluginStreamManager({
             relayUrl: options.relayUrl,
@@ -1192,7 +1222,7 @@ export async function createHerdrSessionSource(
         if (!agentsByPane.has(record.paneId)) {
             throw new Error(`${record.agentName} is not ready for prompts.`);
         }
-        await client.call('agent.prompt', { target: record.paneId, text });
+        await promptHerdrAgent(client, record, text);
     }
     async function sendSessionKeys(sessionId: string, keys: string[]): Promise<void> {
         const record = await resolvePane(sessionId);
--- a/apps/host/src/agent/infrastructure/layoutSelfCheck.ts
+++ b/apps/host/src/agent/infrastructure/layoutSelfCheck.ts
@@ -13,7 +13,7 @@ import {
 import { lifecycleReasonForObservation } from '../domain/lifecycle.js';
 import { IdentityStore, parseTaskTitle } from './identity.js';
 import { RealtimeCodingCoordinator } from './realtimeCoordinator.js';
-import { closeAgentRoute, closeExactPane, closeExactTab, closeExactWorkspace, sendKeysToLiveAgent } from './herdrSessionSource.js';
+import { closeAgentRoute, closeExactPane, closeExactTab, closeExactWorkspace, promptHerdrAgent, sendKeysToLiveAgent } from './herdrSessionSource.js';
 import { createConnection } from 'node:net';
 import { mkdtempSync, writeFileSync } from 'node:fs';
 import { tmpdir } from 'node:os';
@@ -316,6 +316,24 @@ async function demo(): Promise<void> {
     const socketPath = join(socketDir, 'realtime-coding.sock');
     const prompts: string[] = [];
     const sentKeys: Array<{ sessionId: string; keys: string[] }> = [];
+    const herdrPromptClient = {
+        call: async <T>(_method: string, params: Record<string, unknown> = {}): Promise<T> => {
+            const prompt = String(params.text ?? '');
+            if (prompt.startsWith('malformed')) return { type: 'agent_prompted', agent: { pane_id: 'w1:p1' } } as T;
+            return {
+                type: 'agent_prompted',
+                agent: {
+                    terminal_id: 'terminal-one',
+                    agent_status: 'idle',
+                    workspace_id: 'w1',
+                    tab_id: 'w1:t1',
+                    pane_id: prompt.startsWith('wrong pane') ? 'w1:p2' : 'w1:p1',
+                    focused: false,
+                    revision: 1,
+                },
+            } as T;
+        },
+    };
     const coordinatorAgents = [
         { sessionId: 'pp_john', cwd: '/repo', displayName: 'John', taskTitle: 'Harden audio', kind: 'pi', status: 'idle' },
         { sessionId: 'pp_maria', cwd: '/repo', displayName: 'Maria', taskTitle: 'Ship settings', kind: 'pi', status: 'working' },
@@ -324,7 +342,10 @@ async function demo(): Promise<void> {
         list: async () => coordinatorAgents,
         activity: async () => [],
         start: async () => ({ accepted: false }),
-        prompt: async (_sessionId, text) => { prompts.push(text); },
+        prompt: async (_sessionId, text) => {
+            await promptHerdrAgent(herdrPromptClient, { paneId: 'w1:p1', agentName: 'John' }, text);
+            prompts.push(text);
+        },
         sendKeys: async (sessionId, keys) => { sentKeys.push({ sessionId, keys }); },
         read: async () => ({ text: '', truncated: false }),
         status: async () => 'idle',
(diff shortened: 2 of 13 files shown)


umeranjum17 (commented):
**REQUEST CHANGES** — reviewed at `229867e2021430cb5fcee7e7a2f3409a586dc4ae` against `main` and the incident requirements.

The core of the fix is sound. Requirements 1–4 hold at the host boundary, `agent.prompt` has exactly one call site so the receipt check covers every caller, and the receipt validation matches the published herdr schema field-for-field. Four items block, all in requirements 5 and 6.

---

## Blocking

### 1. `safeSemanticName` redacts before `.normalize('NFKC')` — secrets land in the journal
`apps/host/src/diagnostics/infrastructure/journal.ts:73-86`

Every redaction regex runs on the raw string; NFKC at line 82 then folds compatibility forms back into plaintext ASCII, and that is what is persisted. Verified by running the shipped function:

```
ｓｋ－LIVEKEY123456789      -> "sk-LIVEKEY123456789"
ｔｏｋｅｎ＝supersecretvalue  -> "token=supersecretvalue"
Ｂｅａｒｅｒ AAAAAAAAAAAAAAAA -> "Bearer AAAAAAAAAAAAAAAA"
```

ASCII equivalents redact correctly, so the flow test does not catch it. Reachable via this PR's own new branch: `realtimeCoordinator.ts:510-516` forwards **raw** `request.agent` to `promptDiagnostic`, bypassing the only NFKC in the pipeline (`string()`, `realtimeCoordinator.ts:85`); `safeProviderText` (`realtimeCoordinator.ts:214-233`) has no normalize either. `prompt_agent({agent:'ｓｋ－…', text:''})` fails parse on the empty text and writes the cleartext key to `~/.muxr/host/diagnostics.json`.

Fix: make `.normalize('NFKC')` the first operation in the chain.

### 2. Internal-id regex misses most real herdr pane/tab ids
`apps/host/src/diagnostics/infrastructure/journal.ts:79`

`(?:w\d+[A-Za-z]?):(?:p|t)\d+` permits at most one letter after the workspace digits. Checked against live `herdr pane list`:

```
w1:p1    -> [internal reference]     wAG:p9S -> wAG:p9S
w18J:p1  -> [internal reference]     w1AK:p1 -> w1AK:p1
                                     w1BS:p6 -> w1BS:p6
                                     w1BS:t6 -> w1BS:t6
```

Most live ids pass through verbatim. The spec test asserts only on `pp_secret`, which the *other* alternative in the same regex matches, so the test is green while the real-world format leaks.

The same pattern is duplicated at `realtimeCoordinator.ts:223` and `plugins/voice/coordinatorPolicy.mjs:293,306`, so this also affects spoken provider text — and CLAUDE.md forbids speaking internal ids. Please fix all four sites; something like `w[0-9A-Za-z]+:(?:p|t)[0-9A-Za-z]+` covers the observed shapes.

### 3. A prompt that dies in `resolve()` writes no journal event
`apps/host/src/agent/infrastructure/realtimeCoordinator.ts:350`

`queuePrompt` propagates before the first `promptDiagnostic` at `:353` when `handlers.list()` → `refreshSnapshot()` rejects (herdr socket down or timing out). One line lower, the structurally identical `handlers.prompt` failure **is** journaled as `failed` (`:359-360`). An operator running `muxr diagnostics` after "voice said something and the agent got nothing" cannot distinguish a failed prompt from one that was never attempted — which is the exact investigation this journal exists to support. No false success is reported to the provider; this is an observability asymmetry inside the PR's own new code.

### 4. Codex delegation is not fail-closed, it is permanently inert — say so explicitly
`plugins/voice-codex/stream.mjs:216`

`delegate()` is the adapter's only route into muxr agents and passes `{ text: request }` with no `agent`. The Codex session is opened with `delegation: { type: 'client' }` and **no tools array** (`stream.mjs:269-272`), so the provider protocol has no field in which a target name could ever travel, and no `list_agents` to recover with. Every delegation returns `"Which named agent should I prompt? Ask me to list agents."`; the user answers "Beta", Codex re-delegates with "Beta" inside `text` and still no `agent`. The loop never terminates. It is blocked at both layers (`coordinatorPolicy.mjs:267` and independently `realtimeCoordinator.ts:128`), so removing the plugin guard would not restore it.

On `main` this worked, and not by guessing: `resolve(state, undefined)` fell back to `state.activeSessionId`, which for a session-scoped voice stream is the pane the user opened voice on (`herdrSessionSource.ts:1575` → `pluginStreamManager.ts:157-161`) — an unambiguous, host-known target, unlike the incident's silent capture.

If the decision is to remove the capability, that is a reasonable call for a default-disabled experimental adapter (`checkPackageSmoke.mjs:645`), but it needs to be stated as a removal rather than as a conditional fail-closed. Concretely:
- `docs/specs/plugin-primitives.md:68,89` — "fails closed when it cannot supply a semantic target" reads as an edge case; it is 100% of Codex delegations.
- `plugins/voice-codex/stream.mjs:36` — the system prompt still instructs the model to "Delegate coding work to the client".
- The clarification string points at a `list_agents` tool this adapter does not expose.

---

## Non-blocking

**`send_agent_keybinding` still mutates an unnamed active agent** — `realtimeCoordinator.ts:442`. Pre-existing and outside this PR's declared scope, but it is the surviving instance of the implicit-target pattern on the coordinator's other mutation, and `queuePrompt`'s `activate()` (`:363`) now moves the active pointer more often. The receipt names the agent that actually received Escape, so a misroute is audible rather than silent. Worth a follow-up.

**Requirement 1's host proof lives only in `layoutSelfCheck.demo()`** — reverting *only* `parseRequest`'s `string(request.agent,'agent',160)` back to `optionalString` leaves every assertion in the Gemini flow green; it is caught solely by `layoutSelfCheck.ts:363-366`. That check does run in CI (`runSuite.mjs:27`, ungated), so the proof is enforced — this is a note on where it lives, and explains why the expected diagnostics array has four entries rather than five: the plugin guard at `coordinatorPolicy.mjs:267` short-circuits the missing-target case before the socket.

**Peer broker watch item** — `apps/host/src/peer/infrastructure/broker.ts:212` still routes an omitted `agent` to whatever single agent the remote machine has. Unreachable today (`plugins/voice/peerBroker.mjs` has no importers), but the capability is still issued to every `voice.session` plugin.

---

## Verified sound

- **Req 1** — enforcement is genuinely host-side, not plugin-only. `parseRequest` rejects absent/empty/whitespace-only/>160B before `invoke()`; the `CodingRequest` union was narrowed to `agent: string`; `resolve()`'s empty-spoken branch is unreachable from prompt (no empty alias key can exist — `publicAgent` falls back to `'Agent'` and `title()` to `` `${kind} task` ``).
- **Req 2** — rejection returns before both `handlers.prompt` and `activate()`; a rejected prompt does not rebind the capability's active agent.
- **Req 3** — `agent.prompt` has exactly one call site repo-wide (`herdrSessionSource.ts:186`), and the mobile `session.prompt` and remote `peer.remote.prompt` paths both land in it, so the check covers all callers. Not over-strict: the seven validated fields are exactly `AgentInfo.required` in the published herdr schema, `agent_prompted` returns `{type, agent}`, and `herdr agent get <stale pane>` errors before mutating, so a moved pane cannot produce a wrong-pane receipt.
- **Req 4** — `Queued: instruction for X.` exists in one place (`:364`); `replay()` caches the promise so the replay path is byte-identical; no stale "delivered"/"sent" anywhere on the realtime path; all three tool-capable providers interpolate the updated instructions.
- **Req 5 (visibility)** — `muxr diagnostics` (`scripts/cli.mjs:396`) is the only consumer and dumps events verbatim, no allowlist or exhaustive switch; `validState()` does not shape-check individual events so they survive restart; bounded to 512 events / 256KB / 7 days; raw prompt text is never journaled; `provider` is the real plugin id.
- **Req 7** — reverting `queuePrompt` to the old body fails the Gemini flow at spec:180 and :190 and `demo()` at :359/:365; reverting the pane check fails `demo()` at :365. Both checks run in `yarn check`.

umeranjum17 (commented):
**Delta review `229867e2..cc0be81e`** — scoped to the four findings from the previous review.

**No blockers. All four are met**, each verified by running the shipped code rather than reading it. One non-blocking follow-up below.

---

## 1. NFKC precedes every relevant redaction — met

`.normalize('NFKC')` is now the first operation in all four sanitizers on this path: `journal.ts:74`, `realtimeCoordinator.ts:216`, `coordinatorPolicy.mjs:285` (`redactCredentials`, which feeds both `cleanProviderProse` and `safeTail`), and `pluginStreamManager.ts:41`. Extracted each function at `cc0be81e` and ran it:

```
                          journal chain   safeProviderText   cleanProviderProse
ｓｋ－LIVEKEY123456789   [credential redacted]  [credential redacted]  [credential redacted]
ｔｏｋｅｎ＝supersecret   token=[redacted]       token=[redacted]       token=[redacted]
Ｂｅａｒｅｒ AAAA…        Bearer [redacted]      Bearer [redacted]      Bearer [redacted]
ｐｐ＿johnprivate        [internal reference]   [internal reference]   [internal reference]
```

The exploit route I reported — `realtimeCoordinator.ts:510-516` forwarding raw `request.agent` on a parse failure — is closed, because `safeProviderText` now normalizes before its own redaction rather than relying on `string()` upstream.

## 2. Real Herdr routes removed from speech and diagnostics — met

`w[0-9A-Za-z]+:(?:p|t)[0-9A-Za-z]+` applied at `journal.ts:80`, `realtimeCoordinator.ts:224`, `pluginStreamManager.ts:47`, `coordinatorPolicy.mjs:294` and `:306`. Every live id from this machine's `herdr pane list` now redacts through all four chains:

```
wAG:p9S  w1AK:p1  w1BS:t6  w1BH:p1   ->  [internal reference]   (all four sanitizers)
```

## 3. Pre-resolution failures journal `failed` — met

`realtimeCoordinator.ts:350-353` — `resolve()` rejection is caught, journals `outcome: 'failed'` with the requested semantic name and `undefined` → `null` resolved name, then rethrows. The signature also narrowed to `requestedAgent: string` and the dead `?? 'active agent'` fallback is gone.

## 4. Codex fails closed with explicit speakable copy, no unnamed coordinator call — met

`plugins/voice-codex/stream.mjs:213` returns a hardcoded string; the `runCodingTool` import is removed and `grep` confirms no remaining reference anywhere in the adapter, so there is no coordinator call on any path. Copy is delivered via `appendContext(result, id)` on `channel: 'speakable'`, and reads: *"Codex Voice could not queue that instruction. An explicit Agent Name or Task Title is required."*

---

## Focused assertions — load-bearing

Checked each new assertion against the change it would catch on revert:

| Assertion | Catches |
|---|---|
| `providerRefusal(502, 'ＸＡＩ＿ＡＰＩ＿ＫＥＹ＝…')` → `[credential redacted]` (spec:75-77) | NFKC in `redactCredentials` |
| `cleanProviderProse('wAG:p9S w1AK:p1 w1BS:t6', …)` → three `[internal reference]` (spec:78-79) | route regex on the speech path |
| journal `requestedAgentName: 'token=[redacted] API_KEY=[redacted] [path hidden] [internal reference] ×4'` (spec:225) | NFKC **and** route regex in `safeSemanticName`, end-to-end through the real Gemini child process |
| `privateRejected` contains `[internal reference]`, not `wAG:p9S`/`key-private` (spec:181-182) | `safeProviderText` on the spoken clarification |
| `listFails` → `{requestedAgentName:'Beta', resolvedAgentName:null, outcome:'failed'}` (spec:228-234) | the `resolve().catch` in finding 3, **and** that the thrown error's own `token=list-private w1AK:p1` does not leak |
| `expect(prompts).toHaveLength(1)` (spec:186) | zero-mutation across all five prompt calls |
| Codex `delegation.created` → exact speakable string, `not.toMatch(/Queued:\|sent\|delivered/i)` (spec:750-773) | finding 4, over the real WebRTC data channel |

Every one fails if its change is reverted. The extended negative grep at spec:244 now also covers `must not survive`, `key-private`, `list-private`, and all three live route forms.

---

## Non-blocking follow-up

**Four sanitizers outside this path still carry both original bugs** — old route regex, and two of them no NFKC at all:

- `apps/host/src/peer/infrastructure/broker.ts:47` (`safeVoiceOutput`, no NFKC at `:41`)
- `apps/host/src/peer/application/outboundPeerService.ts:28` (`name()`)
- `apps/mobile/sources/conversation/application/realtimeSessionState.ts:49` (`visibleVoiceDetail`, no NFKC at `:42`)
- `apps/mobile/sources/watch/domain/voiceReport.ts:12-13`

Verified by running the two full ones — both leak every fullwidth secret and every live pane/tab id unchanged. My original review named only four sites; the sweep needed eight. That is my miss, not a regression in this delta.

Not blocking, because neither carries realtime prompt speech or diagnostics:

- `safeVoiceOutput` sanitizes remote terminal output for the voice model, but `runPeerTool`/`peerOnlyTools` in `plugins/voice/peerBroker.mjs` still have **no importers**, so that path is unreachable from any shipped adapter.
- `visibleVoiceDetail` is defence-in-depth: `pluginStreamManager.ts:332-334` applies the now-fixed `safeStreamReason` to `realtime.state` detail before it reaches the phone, and pane ids do not originate on the device.
- `voiceReport.ts` Agent Names are gated by `agentNameIsTrusted`'s separate whole-string `/^[\w-]+:[\w-]+$/`, which already matches the real form.

Worth a follow-up so all eight share one pattern; `safeVoiceOutput` is the one to fix first, since it is the only other sanitizer whose output is spoken.

**Also unchanged, as previously noted:** the Codex `PROMPT` at `stream.mjs:35` still says "Delegate coding work to the client", so the model will keep attempting delegation and hearing the refusal. With explicit copy the user is no longer misled about the outcome, so this is a UX polish item rather than a correctness one.

umeranjum17 (commented):
**Rebase review** — `cc0be81e^!` vs `f37736c7^!`, rebased onto `origin/main` `19b196f61addc19d90f017686a7d7dbb4b970573`.

**Approved. No blockers.** Byte/delta only; no re-analysis of the patch content, which was reviewed and approved at `cc0be81e`.

## The patch is unchanged

| Check | Result |
|---|---|
| `git range-diff cc0be81e^..cc0be81e f37736c7^..f37736c7` | `1: cc0be81e = 1: f37736c7` |
| `git patch-id --stable` | both `a373e2cc3a1524d6eb29af4bca047a120bb3cd9d` |
| Byte diff of both patch texts, `index` lines stripped | identical, 931 lines |

Subject and author metadata match. `index` lines are the only difference, which is expected once parent blobs change.

## Rebase is onto the stated base

`f37736c7^` = `19b196f61addc19d90f017686a7d7dbb4b970573` = `origin/main` = `fix(mobile): restore agent provider logos` (#166). Branch tip `origin/fix/realtime-prompt-delivery` is `f37736c7`.

## #166 is inherited, not overridden

- **Zero file overlap.** #166 touches 23 files, all under `apps/mobile/sources/assets/agents/` plus `components/AgentGlyph.tsx`, `agentImageKind.ts`, `agentImageKind.spec.ts`. This PR touches 13 files, all under `apps/host/src/`, `plugins/`, `docs/specs/`. `comm -12` over the two sorted lists is empty.
- **Blob-level identity, not just an empty diff.** `git ls-tree -r` across all 23 #166 paths hashes to `742af7a1a28e46a6d973f0d8d726475be3b48a921ed3915068ca09c2b5a866f5` in both `19b196f6` and `f37736c7`, and all 23 logo files are present in the PR tree — so the empty diff is not a pathspec false negative.
- **No stray inherited content.** `git diff origin/main...f37736c7` is byte-identical to the commit patch, so the PR contributes exactly its 13 files.
- **No conflict residue** (`<<<<<<<` / `=======` / `>>>>>>>`) in any file of the commit.

Nothing in this rebase re-opens the four findings verified at `cc0be81e`; the prompt/privacy patch is bit-for-bit the reviewed one.

## Follow-ups still open (unaffected by the rebase — neither file is in this PR)

- `apps/host/src/peer/infrastructure/broker.ts:47` — `safeVoiceOutput`, old route regex and no NFKC at `:41`. Only other sanitizer whose output is spoken; unreachable today since `runPeerTool`/`peerOnlyTools` in `plugins/voice/peerBroker.mjs` have no importers.
- `apps/mobile/sources/conversation/application/realtimeSessionState.ts:49` — `visibleVoiceDetail`, same two bugs at `:42`; defence-in-depth behind the now-fixed `safeStreamReason` at `pluginStreamManager.ts:332-334`.

PR #168, opened 2026-08-29
refactor: make Herdr agent identity authoritative

umeranjum17:
## Summary
- replace persisted muxr identity copies with an owner-only random Agent Route to Herdr `agent_session` binding
- project current Herdr name, provider, labels, topology, lifecycle, and prompt readiness directly across host, peer, voice, and mobile
- fail closed across generation replacement, disable mobile prompting until the current pane is promptable, and remove pane-title restoration by pane
- address all eight Opus review findings, including partial-event readiness, privacy probes, live tree rows, single pane resolution, and rolling peer compatibility

## Verification
- `yarn run check` — 34/34 passed after review fixes
- targeted Opus regressions — 40 tests passed plus layout and lifecycle self-checks
- live isolated Herdr backend loop — passed
- API 36 x86_64 release built with `scripts/buildAndroidLocal.sh`, installed on emulator, launched, and inspected

## Artifacts
- `muxr-api36-x86-release.apk` SHA-256 `c0b70779f6397c34e96ee28bf775a7c35551c3b2e3a3d7db521411b5203ede0e`
- API 36 surface screenshot captured after final install

No merge, version, or publish action performed.


umeranjum17 (commented):
## Review verdict: REQUEST CHANGES

Strict correctness + security review of `fix/agent-readiness-truth` against `origin/main` (merge-base `649a941d`, which already contains merged #167). Host, mobile, and contract `tsc --noEmit` are all clean. Eight actionable findings.

---

## High

### 1. `apps/host/src/agent/infrastructure/herdrSessionSource.ts:992-993` — pane events clobber `interactive_ready`/`launch_pending` to `false`, making ready agents transiently un-promptable

The event merge unconditionally overwrites both readiness flags from the event payload:

```ts
interactive_ready: (pane as AgentRecord).interactive_ready === true,
launch_pending: (pane as AgentRecord).launch_pending === true,
```

`agent_status` on the very next line *is* preserved from cache (`...(current?.agent_status === undefined ? {} : ...)`); readiness is not. Only `agent.wait` results and `session.snapshot` records carry these flags anywhere in the repo (they appear at `:176-177`, `:557-558`, `:971`, `:992-993` and nowhere else) — a `pane.agent.status.changed` payload that includes `agent`/`name`/`agent_session` but not `interactive_ready` sets it to `false`.

**Failure:** a working agent goes `working -> idle`. The handler at `:1010-1016` merges (readiness -> false), then calls `emitState` **in the same tick**, which publishes `status.update` with `promptable: false` and fires `onAgentReadinessDiagnostic('starting', false)`. `scheduleResnapshot()` only repairs this 500 ms later. In that window `promptSession` rejects with `agent-not-ready`, and mobile disables the composer (`TerminalScreen.tsx:672` `editable={panePromptable}`) and refuses sends (`sync.ts:643`). This flickers on every status transition — the most-fired event in the system.

**Fix:** preserve cached readiness when the event omits it, mirroring the `agent_status` line directly below.

### 2. `apps/relay/src/push/infrastructure/push.ts:66-68` and `apps/host/src/agent/infrastructure/lifecycleStore.ts:26-27` — path-leak rejection is bypassable by one leading whitespace character

The old code normalized / collapsed / **trimmed** before testing; the new code tests the raw string. Verified:

| Task Title | old | new |
|---|---|---|
| `" /Users/umer/clients/acme/secret-plan"` | rejected | **accepted** |
| `"\u00a0C:\\Users\\umer\\payroll"` | rejected | **accepted** |

**Failure:** a pane/tab label with a leading space (or NBSP — NFKC normalization was also dropped) containing an absolute path now passes both the host lifecycle store and the relay push parser, and is rendered into a push notification body and persisted into `lifecycle.json`. This is exactly the leak the `^(?:\/|[A-Za-z]:\\)` guard exists to stop.

**Fix:** trim (and NFKC-normalize) before testing, or reject on leading whitespace.

---

## Medium

### 3. `apps/host/src/agent/infrastructure/realtimeCoordinator.ts:184` — Agent Name is no longer sanitized before it reaches the third-party voice provider

```ts
displayName: agent.displayName,   // was: normalizeAgentName(cleanHuman(agent.displayName, 'Agent', 80))
```

`publicAgent()` is the sanitization boundary for provider-facing text. Its output flows raw into provider strings at `:406` (`${agent.displayName} — ...`), `:440` (`Confirmed: ${created.displayName} ...`), `:322`/`:326` (clarification prompts) and every `Confirmed:` reply. The 80-char cap and the ``[\u0000-\u001F\u007F<>`{}]`` strip are gone.

The inconsistency is self-evident in the same file: `:418` still applies `cleanHuman(lifecycleEventAgentName(event), 'Agent', 80)`, and `:340` still applies `safeProviderText(...)` to the requested name.

**Failure:** `herdr agent rename <pane> "<10 KB of text with </instructions> and backticks>"` is forwarded verbatim to the realtime provider on the next `list` tool call.

**Fix:** re-apply `cleanHuman(..., 80)` inside `publicAgent`.

### 4. `apps/host/src/peer/infrastructure/store.ts:105` — `machineAlias` validation deleted along with `agentNames`, but the field is still live

The removal took out two predicates, not one. `machineAlias` remains on `StoredPeerRelationship` (`store.ts:27`) and is read at `outboundPeerService.ts:106` (`key(entry.machineAlias ?? '')`) and `:288` (`relationship.machineAlias?.trim()`).

**Failure:** a `peers.json` whose relationship carries `machineAlias: 123` now loads. `?.` guards only null/undefined, so `:288` throws `TypeError: relationship.machineAlias.trim is not a function` on `ensureMachineAlias`, which every peer request path calls (`:56`, `:68`, `:72`, `:217`, `:230`, `:244`) — all peer operations fail until the file is hand-edited.

**Fix:** restore the `machineAlias` predicate.

### 5. `apps/mobile/sources/herd/application/sessionUtils.ts:120` — every Herdr session row renders as "New chat"

```ts
return session.metadata?.summary?.text?.trim() || t('session.newChat');
```

But `summary` is no longer written for Herdr sessions: `sessionMapping.ts` dropped `summary: { text: fields.taskTitle, ... }`, and `agent.ts` `withoutCopiedHerdrIdentity` actively strips `summary` from any metadata whose `client.id === 'herdr'` (so pre-upgrade rows are cleaned too). `metadata.taskTitle` / `agentName` were removed from `MetadataSchema` entirely.

**Failure:** `buildSessionRowData` (`storage.ts:91`) yields `name: "New chat"` for every agent; `SessionsList.tsx:400` and `ActiveSessionsGroupCompact.tsx:283` render it, `MainView.tsx:528` mounts `SessionItem`. Attention/inbox rows (`storage.ts:620`) show "New chat" instead of the Agent Name; `machine/[id].tsx:432` and `recent.tsx:192` too. `filterSessionList` (`sessionRowPresentation.ts:60-66`) matches on `session.name` + `session.subtitle`, and `getSessionSubtitle` now returns `''` (`sessionUtils.ts:157`) — so **search by Agent Name or Task Title is dead** on these surfaces.

**Fix:** these rows need to read the tree (as `sync.ts:56 currentAgentName` does) rather than a field nothing populates.

### 6. `apps/host/src/agent/infrastructure/terminalManager.ts:59` and `:84` — `resolvePane` is invoked twice per attach, and the two results are not guaranteed equal

`main.ts:611-615` implements `resolvePane` as `source.open({...})`, and `open` (`herdrSessionSource.ts:1735`) calls `resolvePane` -> `refreshSnapshot()` -> a full herdr `session.snapshot` RPC plus `syncDiscovery()`. `attach()` resolves at `:59` to key `controlQueues`, then the queued `attachNow()` resolves again at `:84`. The old code did one synchronous `identity.get()` lookup and reused the value.

**Failure:** (a) two full snapshots + two discovery reconciles per terminal attach, on a path that runs per reconnect; (b) if a `pane.moved` lands between them, `controlQueues` is keyed on the stale pane id while `:241` (`current.paneId === paneId`) dedupes on the new one — the previous controller for that pane is not closed, defeating the "Linearize same-pane control requests" guarantee the queue exists for.

**Fix:** resolve once in `attach` and thread the id into `attachNow`.

### 7. `apps/host/src/peer/application/outboundPeerService.ts:303-311` — a delivered peer prompt is reported as failed when the post-hoc name lookup misses

`currentAgentName` issues an extra `session.list` **after** the mutation and throws `agent-unavailable` when the session is absent:

```ts
const agentName = sessions.find((session) => session.id === sessionId)?.agentName;
if (agentName === undefined) throw operationError('That agent is no longer available.', 'agent-unavailable');
```

Called at `:230` after `session.prompt` already succeeded, and at `:246` after `session.start` already created the agent.

**Failure:** the prompt is queued, the agent finishes and its pane closes within that round-trip, the lookup misses, and `runStored`'s catch marks the mutation `delivered` / `ok:false` — the user is told the prompt failed when it was delivered.

**Second-order:** `session.list` on a peer running pre-#168 host code returns `displayName`, not `agentName`, so **every** `peer.remote.read` / `status` / `watch` / `prompt` / `start` against an older peer throws `agent-unavailable`, and `peer.remote.list` (`:50-51`) returns an empty session set.

**Fix:** capture the name from the response you already have, or degrade to a fallback instead of throwing.

---

## Low

### 8. `apps/mobile/sources/herd/domain/agentPresentation.ts:27` — `agentName` can now be `''`, defeating the downstream fallback chain

```ts
agentName: agentKind === undefined ? 'Shell' : pane?.agentName ?? '',
```

Previously this returned a guaranteed-nonempty value (`normalizeAgentName` -> `'Agent'`). `TerminalScreen.tsx:369` is the only caller passing a `session`: when the herdr tree has not loaded, `pane` is `undefined` but `agentKind` resolves from `session.metadata.agentKind`, so `agentName` is `''`. `sessionContextTitle` (`TerminalLink.ts:46`) uses `??`, so `''` short-circuits past both the `agentKind` and `'session'` fallbacks and the header renders blank.

**Related:** `tabs` is seeded `[]` and `loadSiblings` swallows failures (`TerminalScreen.tsx:203` `.catch(() => undefined)`) with no retry, so one failed `herdr.tree` at mount leaves the composer disabled until `session.metadata.promptable` happens to flip.

---

## Checked, no finding

- **Route / generation fencing** — `agentRouteStore.ts` key covers all four `agent_session` fields; `reconcile` deletes from a `Map` while iterating `values()`, which is safe in JS, and it clears both indices; `bind` is collision-checked; `0o600` + atomic write hold, and `layoutSelfCheck.ts:172-181` asserts the restart case.
- **Stale route removal** — `refreshSnapshot` throws on RPC failure rather than passing an empty `liveSessions`, so a herdr outage cannot mass-wipe bindings.
- **#167 regressions** — all four guarantees survive: `promptHerdrAgent` still validates `agent.pane_id === target.paneId` (`:224`), `namedAgent` keeps the target non-empty and non-`pp_`, `promptPromptableHerdrAgent` fails closed, and `providerRefusal.spec.mjs` only renamed a fixture field.
- **Mobile lifecycle** — `paneStatus` (`herd.ts:75-83`) is a faithful inline of the deleted `Agent.lifecycle()`.
- **Pane-titler generation safety** — `log()` does its own `mkdirSync`, no unused imports remain, `formatTitle` bounds output to 4 words / 40 chars before `execFileSync` (argv, no shell), and `title.spec.mjs` genuinely fails if the restore branch returns.
- **Shell behaviour** — shell panes were already un-promptable on `main` (`promptSession` required `agentsByPane.has(paneId)`), so gating the composer is not a regression.
- **Peer replay** — duplicate peer prompts are prevented by remote-side `operationId` receipts.

umeranjum17 (commented):
## Review verdict: APPROVE

Delta review `4ffa428f` → `6a529353` (21 files, +204/−66). All eight findings from [the previous review](https://github.com/umeranjum17/muxr/pull/168#pullrequestreview-5057915257) are resolved. `tsc --noEmit` clean on host, mobile, contract, and relay. No regression found in the delta.

---

## Finding-by-finding verification

### 1. Readiness clobber — FIXED
`herdrSessionSource.ts:183-188` extracts `mergeHerdrAgentEvent(current, incoming)` = `{ ...current, ...incoming }`, and `:996-1001` uses it in place of the two forced `=== true` coercions. A partial `pane.agent.status.changed` payload no longer resets `interactive_ready`/`launch_pending`, while an explicit replacement value is still honoured. This restores the same merge shape `agent_status` uses on the following line.
Pinned by `layoutSelfCheck.ts:113-122` — asserts both directions (partial preserves, explicit replaces).

### 2. Path-leak bypass via leading whitespace — FIXED
`lifecycleStore.ts:26-29` and `push.ts:67-69` now derive a `privacyProbe = value.normalize('NFKC').trimStart()` and run the `^(?:\/|[A-Za-z]:\\)` / secret-assignment test against it, while still storing the exact Herdr presentation value. Re-ran the original bypass vectors against the new predicate:

| Task Title | result |
|---|---|
| `" /Users/umer/clients/acme/secret-plan"` | rejected |
| `"\u00a0C:\\Users\\umer\\payroll"` | rejected |
| `"\t/etc/shadow"` | rejected |
| `"Restore full-size terminal cards"` | accepted |
| `"Review monitoring"` | accepted |

NFKC folds NBSP to U+0020 before `trimStart`, so the non-ASCII vector closes too, with no false positives on legitimate titles.
Pinned by `watchStores.selfCheck.ts:82-90` (host) and `persist.integration.test.ts:227-234` (relay).

### 3. Unsanitized Agent Name to the voice provider — FIXED
`realtimeCoordinator.ts:184` restores `cleanHuman(agent.displayName, 'Agent', 80)`, re-establishing the 80-char cap and the ``[\u0000-\u001F\u007F<>`{}]`` strip at the `publicAgent()` boundary — consistent again with `:418` and `:340`. Dropping `normalizeAgentName` is fine here: `namedAgent()` already excludes `pp_`/`pph_` names from routes upstream.
Pinned end-to-end by `providerRefusal.spec.mjs:273` + `:431-433` — an agent named `Unsafe<script>` reaches the provider as `Unsafe script`, and `<script>` is asserted absent.

### 4. `machineAlias` validation — FIXED
`store.ts:106` restores the predicate that was collaterally removed with `agentNames`.
Pinned by `peerFlow.test.ts:178-198` — a `peers.json` carrying `machineAlias: 123` now throws `unsupported peer schema` instead of loading and later `TypeError`-ing in `ensureMachineAlias`.

### 5. "New chat" on every Herdr row — FIXED
`getSessionName`/`getSessionSubtitle` (`sessionUtils.ts:120-124`, `:158-161`) take an optional `HerdrTreePane` and read `pane.taskTitle ?? pane.agentName` / `pane.agentName`. New `herdrPaneForSession` (`agentPresentation.ts:12-24`) resolves it, and **all five** call sites now pass it: `storage.ts:87-92` (`buildSessionRowData`), `storage.ts:630-636` (`useAttentionRows`), `machine/[id].tsx:433-434`, `recent.tsx:193-195`, `useLandWorktree.ts:24`.

The important second half is also handled: `storage.ts:327-333` makes `applyHerdrTree` rebuild `sessionListViewData`, so rows built before the tree arrived are refreshed rather than stuck on the fallback. `filterSessionList` therefore matches Task Title (`name`) and Agent Name (`subtitle`) again — search is restored. `taskTitle` cannot arrive as `''` (the host's `taskTitleForSession` maps empty to `undefined`), so `??` is the right operator here.
Pinned by `sessionSync.integration.spec.ts:354-365` — after `applyHerdrTree`, the active row asserts `name: 'Stabilizing realtime voice'`, `subtitle: 'Maria'`.

### 6. Double `resolvePane` per attach — FIXED
`terminalManager.ts:59` resolves once; `attachNow` now takes `paneId` as a parameter (`:75-83`) and both call sites (`:60` observe, `:65` queued control) pass the same value. The `controlQueues` key and the `:240` dedup comparison can no longer disagree, and the per-attach snapshot cost is halved. Confirmed `:60` and `:65` are the only callers.
Pinned by `terminalManager.test.ts:116` — `expect(resolvePane).toHaveBeenCalledTimes(4)` for four attaches.

### 7. Delivered peer prompt reported as failed — FIXED
Three separate corrections:
- `currentAgentName` (`outboundPeerService.ts:313-322`) no longer throws; it degrades to `'Agent'`.
- The lookup now runs **before** the mutation for `peer.remote.watch` (`:219`) and `peer.remote.prompt` (`:236`), so a post-mutation miss can no longer turn a delivered prompt into a failure.
- `peer.remote.start` (`:254`) reads the name straight off `snapshot.info` instead of issuing an extra `session.list`.

The cross-version break is closed too: `currentSessionAgentName` (`:37-41`) falls back to a legacy `displayName`, so peers running pre-#168 hosts keep working.
Pinned by `peerFlow.test.ts:360-363` (legacy `displayName` peer lists correctly) and `:531-539` (`dropSessionsAfterPrompt` — the session vanishes during the prompt round-trip and the result is still `delivered: true, agentName: 'iOS builder'`).

### 8. Empty-string `agentName` — FIXED
`agentPresentation.ts:40` returns `pane?.agentName ?? 'Agent'`, restoring a guaranteed-nonempty value so `sessionContextTitle`'s `??` chain (`TerminalLink.ts:46`) no longer short-circuits on `''` and the terminal header renders.
Pinned by `liveTerminalOrder.spec.ts:94`.

---

## Regression sweep on the delta

- **Typecheck:** `apps/host`, `apps/mobile`, `packages/contract`, `apps/relay` — all clean, no output.
- **`mergeHerdrAgentEvent`** reproduces `main`'s pre-#168 merge shape exactly (spread + forced `pane_id` + preserved `agent_status`), so it reintroduces nothing that was previously fixed.
- **No import cycle:** `storage.ts:36` imports `herdrPaneForSession` from the deep path `@/herd/domain/agentPresentation` rather than the `@/herd` barrel (which re-enters `@/catalog`); `sessionUtils.ts:7` imports only a type.
- **Hook rules:** `useHerdrTree()` is called unconditionally at the top of `useAttentionRows`, `useLandWorktree`, `recent.tsx`, and `machine/[id].tsx`; `workspaces` was correctly added to the `useMemo`/`useCallback` dependency arrays (`storage.ts:640`, `recent.tsx:236`).
- **Zustand `set` is a shallow merge**, so the new `sessionListViewData` key in `applyHerdrTree` does not drop sibling state.
- **No test was weakened.** Every changed assertion tightened, and seven new ones were added that fail if the corresponding fix is reverted.

## One non-blocking note (not a finding)

`storage.ts:633` — when an attention entry's session still exists but its pane has left the tree, the row falls back to `t('session.newChat')` rather than the `pane?.taskTitle ?? pane?.agentName ?? 'Agent'` chain used on the adjacent `session === undefined` branch. Narrow (a session whose pane has closed while the attention row survives) and cosmetic. Worth a one-line tidy whenever this file is next touched; it does not hold up the merge.

Nice work — the fixes are minimal, each landed at the root rather than the call site, and each arrived with a check that pins it.

PR #192, opened 2026-08-31
fix: make starting an agent from the Dock smooth

umeranjum17:
## Symptom
Starting a new agent from the Dock popped **"Agent is not ready yet"** every time, while the agent was visibly up at its prompt.

## Root cause 1 — the popup
`sync.sendMessage` refused whenever *cached* session metadata said the pane was not promptable:

```ts
if (storage.getState().sessions[sessionId]?.metadata?.promptable !== true) throw ...
```

A freshly started agent is not promptable for the first few seconds. Measured against the real host, Codex publishes `promptable=false` until ~5s:

```
+1557ms  promptable=false lifecycle=unknown
+4006ms  promptable=false lifecycle=unknown
+5132ms  promptable=true  lifecycle=idle
```

The Dock sends its first prompt inside that window, so the client threw before the host ever saw it. The host already waits for a starting agent to become promptable — this check only contradicted it and ate the prompt.

## Root cause 2 — the dead Dock
Removing that check exposed the next problem: `startSessionFromDraft` navigated only after `startAgentFromDock` fully resolved, and that now includes the host waiting for the agent. So Send sat on an unresponsive Dock for the whole boot.

Navigation now fires as soon as the route exists; the prompt lands behind the terminal that is already on screen.

## Emulator verification
Codex from the Dock, project `~`, prompt typed:

| | before | after |
|---|---|---|
| popup | "Agent is not ready yet" | none |
| navigate | ~3s, after prompt delivery | ~1.5s, on route |
| composer | flashes "Agent is not ready yet" | "Type a prompt…" |
| prompt | dropped | lands |

Verified on a live emulator paired to the real host: `hello` reached the agent, header read `codex/yak`, `~` resolved correctly.

Also fixes the worktree handoff, which prompts a just-created agent through the same path.

## Test plan
- [x] `tsc --build` clean
- [x] `sessionSync.integration.spec.ts` + `muxrClient.spec.ts` pass
- [x] Emulator: start Codex from the Dock with a prompt — no popup, prompt lands, immediate navigation

### DIFF
--- a/apps/mobile/sources/catalog/application/sync.ts
+++ b/apps/mobile/sources/catalog/application/sync.ts
@@ -658,11 +658,9 @@ class MuxrSync {
     async sendMessage(sessionId: string, text: string, options?: SendMessageOptions): Promise<void> {
         const previews = options?.attachments ?? [];
         if (text.trim().length === 0 && previews.length === 0) return;
-        if (storage.getState().sessions[sessionId]?.metadata?.promptable !== true) {
-            const error = Object.assign(new Error('Agent is not ready yet'), { code: 'agent-not-ready' });
-            recordTrackedRpc('session.prompt', { ok: false, error }, 0);
-            throw error;
-        }
+        // Readiness is the host's call: it waits for a starting agent to accept
+        // the prompt. Refusing here on cached metadata drops the first prompt
+        // during the seconds before the tree reports the agent promptable.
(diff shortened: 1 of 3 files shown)


umeranjum17:
Superseded by #194, which carries these two commits plus the terminal composer fix, the blank start and the icon changes.

PR #193, opened 2026-08-31
fix: keep the send button lifecycle colour, brighten only idle

umeranjum17:
## Summary
0.1.25 fixed the dull idle send button by painting every state with the accent. That was too broad — it threw away the working blue and done green, which were the good part.

Idle is the only state that reads wrong: `agentStatusColor` has no `idle` case, so it falls through to `status.disconnected` grey and a ready agent looks dead. Give idle the accent and leave every other lifecycle colour alone.

| lifecycle | colour |
|---|---|
| working / starting | `status.working` blue |
| done | `status.done` green |
| blocked / failed | `status.error` red |
| **idle** | **accent (was disconnected grey)** |

The header status text keeps the full lifecycle colour including grey for idle, which is correct there.

## Test plan
- [x] `tsc --build` clean
- [x] Emulator: done agent shows the green filled arrow again
- [ ] Idle agent with text shows the bright arrow


umeranjum17:
Superseded by #194, which includes this send-button colour fix.

Issue #197, opened 2026-08-31
Stuch installation


SalamiYummi:
### What happened

my installation seems to get stuck and does not continue. 

System info:
OS: Linux Mint 22.3 x86_64 
Kernel: 7.0.0-30-generic 

terminal log

```
muxr
◆ muxr 0.1.24
  ● this machine is not set up yet — choose Set up this machine

◆ What would you like to do?
  ●  Set up this computer 
      install prerequisites, choose a connection, and pair your phone
  ○ Advanced setup
      host a shared relay or connect this computer to one running elsewhere
  ○ Show commands
      print the non-interactive command reference
  ○ Quit
  esc back · ctrl-c quit
◆ Check this machine
  … Inspecting Herdr, agents, and networking
◆ Checking this machine
  ● Herdr herdr 0.8.2
  ● Herdr server running
  ● Agent integrations 3 ready — pi, codex, hermes
  ● Tailscale connected — 100.79.52.4
  ○ Cloudflare Tunnel not installed — https://github.com/cloudflare/cloudflared/releases

◆ Choose how your phone connects
◆ muxr runs on this computer
◆ How should your phone connect?
  1. Tailscale · recommended
     private connection · works from anywhere · nothing exposed publicly
  2. Cloudflare tunnel (unavailable)
     not installed — https://github.com/cloudflare/cloudflared/releases
  3. LAN
     works now · phone and computer must use the same wifi
  4. Your own server
     an always-on relay you run · works from anywhere
  Choose 1-4 (1), or b to cancel setup: 1
◆ Choose app access
◆ Host the browser client too?
  1. Native app only
     do not expose the browser client
  2. Host the browser client
     serve it over the selected HTTPS/WSS connection
  Choose 1-2 (1), or b to cancel setup: 1
◆ Choose what to pair
◆ Connect agents and add-ons
◆ Connect your coding agents (3 detected)?
  1. Connect coding agents
     install lifecycle detection so their status stays current
  2. Leave integrations unchanged
     do not change coding-agent lifecycle integrations
  Choose 1-2 (1), or b to cancel setup: 1
◆ Review setup
  │
  │ Connection: Tailscale Serve on local port 8792
  │ Herdr: adopt existing installation and ensure its server is running
  │ Bundled plugins: link the public muxr plugins into Herdr
  │ Agent integrations: sync detected lifecycle providers; leave agent prompt files unchanged
  │ Optional add-ons: none
  │ Browser client: off
  │ Pairing: phone
[...]

PR #199, opened 2026-08-31
fix: detect unavailable Tailscale Serve before setup

umeranjum17:
## Summary

- preserve the real disabled-tailnet notice and its Tailscale admin link, even when `tailscale serve` prints it and then blocks
- distinguish proven disabled Serve from inconclusive timeouts, permission failures, and malformed output
- keep Tailscale recommended when preflight is inconclusive; the bounded Apply remains authoritative
- leave occupied or disabled Serve unchanged and offer direct Tailscale as the safe fallback
- bound sibling readiness paths for the managed relay, service status, remote enrollment, and local token provisioning
- stop newly spawned relay processes when readiness expires and retain timeout details from Herdr checks

Follow-up to #197 and #198. This PR is deliberately limited to failure handling; the onboarding redesign is in a separate stacked PR.

## Safety

- preflight is read-only
- no Tailscale Serve mutation occurs before **Apply setup**
- existing Serve ownership, muxr identity, grants, and pairings remain untouched on failure
- every new external wait has a deadline

## Verification

- `node scripts/diagnostics/application/checkTailscaleIngress.mjs` — covers available, occupied, disabled, malformed, direct, apply, and cleanup flows; the blocking disabled-tailnet fixture exits within its bound
- `mise x node@22.23.2 -- yarn build`
- existing PR checks: CI suite and CodeQL passed at the previous head; they will rerun for the updated head

## Local-suite note

A prior full run was 32/34 locally. The two live-Herdr checks fail identically on `main` because the installed Herdr reports `Agent close plugin RPC is unavailable or changed`; one relay startup also flaked once. The focused changed-flow checks pass.

### DIFF
--- a/docs/SELF-HOSTING.md
+++ b/docs/SELF-HOSTING.md
@@ -52,6 +52,8 @@ are:
 | `--tailscale-direct` | Rollback path using the tailnet IP directly. |
 | *(choose Local network)* | LAN address. Phone must be on the same network. |
 
+Before recommending Serve, the wizard checks that it is available and not already owned. If Serve is disabled for the tailnet, muxr shows Tailscale's enablement link and offers direct Tailscale or LAN instead of waiting indefinitely.
+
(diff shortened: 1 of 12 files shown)


umeranjum17:

PR #201, opened 2026-09-01
feat(mobile): improve file and diff navigation

umeranjum17:
## Summary
- Shared 44dp path chrome, status/+− chips, and prev/next file review on the native viewer without merging the two file mounts.
- Changes resolve from the repository root so a session cwd below the repo still opens the right file; hostile paths are single-quoted for `sessionBash`.
- Stacked on `feat/plugin-herdr-actions` (`c23216a5`). Does not restore Runbook, Ports, or Browser Preview.

## Test plan
- [ ] Changes from a subdirectory cwd opens the correct absolute path
- [ ] Prev/next walks the recorded file list; one Back leaves the review
- [ ] Deep breadcrumbs collapse to first / … / parent / leaf and copy the full path
- [ ] Deleted and hostile filenames show a diff without shell interpolation
- [ ] Commit rail disambiguates duplicate basenames and shows +/− counts

Made with [Cursor](https://cursor.com)

### DIFF
--- a/apps/mobile/sources/app/(app)/_layout.tsx
+++ b/apps/mobile/sources/app/(app)/_layout.tsx
@@ -102,6 +102,14 @@ export default function RootLayout() {
                 }}
             />
             <Stack.Screen
+                name="session/[id]/file"
+                options={{
+                    headerShown: true,
+                    headerTitle: t('common.fileViewer'),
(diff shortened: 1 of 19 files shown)


umeranjum17:

PR #205, opened 2026-09-02
fix(recovery): replace stale routes and restore graphics

umeranjum17:
Generic Tools, a shared document viewer, cross-platform Herdr graphics — and the measurement rig that turned "the phone feels slow with a big herd" into numbers, then a fix.

> Shipping via the consolidated release PR #209 (0.1.26).

## What landed

| Area | Change |
|---|---|
| Tools | generic Tools list driven by Herdr actions, not a hardcoded catalog |
| Viewer | one document viewer shared by files, diffs and attachments |
| Graphics | Kitty images a program writes to its own PTY now reach the phone, on Android and web |
| Perf | `perf/` — a release performance gate and a capacity sweep, both runnable before a release |
| Host | inline Kitty forwarding in the graphics bridge; `HERDR_SOCKET_PATH` so a harness can point the host at its own Herdr |

## Graphics: why the viewport was blank

Herdr 0.8.2 delivers a program's own Kitty images **inline in the type-2 app stream**, positioned by the cursor. The bridge only handled the type-13 leased-file transport, which Herdr never sends for program images, so text streamed and the image area stayed empty forever.

```mermaid
sequenceDiagram
    participant P as program in a pane
    participant H as Herdr 0.8.2
    participant B as host graphics bridge
    participant A as phone
    P->>H: ESC_G a=t,f=32 ... (APC image)
    H->>B: type-2 app stream, inline, at the cursor
    Note over B: before: only type-13 graphics-file handled<br/>inline stream dropped
    B--xA: no pixels, forever
    Note over B: after: inlineKitty scans the stream,<br/>joins chunked transmissions,<br/>routes placements by pane rect
    B->>A: terminal.frame graphics:true
```

Three things turned out to be load-bearing: an image is learned even with **no phone attached** (programs transmit long before anyone connects), unchanged placements are dropped (a repainting producer emits ~30/s), and every forwarded frame must be **self-contained** because the phone's write pump keeps only the newest graphics frame.

## The performance rig

The gate runs our own stack for real and fakes only Herdr, which is third party:

```mermaid
flowchart LR
    subgraph device["Android emulator"]
        APK["release APK under test"]
    end
    subgraph ours["our code, real builds"]
        RELAY["apps/relay/dist"]
        HOST["apps/host/dist"]
    end
    subgraph fake["perf/fake-herdr"]
        CTL["JSON-RPC control socket"]
        CLI["HERDR_BIN shim"]
        GFX["protocol-20 graphics socket"]
    end
    APK <-->|"E2EE over adb reverse"| RELAY
    RELAY <--> HOST
    HOST --> CTL
    HOST --> CLI
    HOST --> GFX
```

`yarn perf` installs the release APK, stands up a relay and host in a scratch directory on kernel-picked ports, pairs over the real E2EE path, drives Maestro flows, then reads JS-thread busy time from `/proc/<pid>/task/<tid>/stat`, frames from `dumpsys gfxinfo`, memory from `meminfo`, update-depth errors from logcat, and the host's own request journal. `adb shell top -H` is used nowhere: it reports a thread's *lifetime* average, which once made a saturating build and a healthy one measure identically.

## Capacity, then the fix

The sweep found the ceiling was the phone, never host RAM — host plus relay stay at 250-330 MB across the whole ramp, about 0.4 MB per pane.

| Herd | Phone JS thread | Host + relay RSS |
|---|---|---|
| 10 panes / 2 agents | 13.3% | 246 MB |
| 30 / 6 | 38.4% | 294 MB |
| 45 / 9 | 56.2% | 303 MB |
| 60 / 12 | 75.9% | 325 MB |
| 100 / 30 | 91-97.6% | 330 MB |

Attribution came from a temporary per-second counter on the frame path: **190 frames/s** at 100/30, and a full second of wall time per second inside the handler. Three costs, all invisible work:

1. **Every decrypted frame awaited a disk write.** `persistReplay` serialized the whole replay cache — one entry per session stream, a hundred of them — and awaited `AsyncStorage` inside `open()`. Now a trailing write. This was the memory climb, not the CPU.
2. **`terminalTitle` was pushed in every `session.updated`**, animated several times a second, rendered by nothing. Removed from `SessionInfo` with the coalescing timers that only existed to throttle it; `herdr.tree` still carries it.
3. **The same lifecycle event was republished** — same `eventId` — whenever only the task title moved, with a journal write each time.

| at 100 panes / 30 agents | before | after |
|---|---|---|
| Phone JS thread | 91-97.6% | **36-43%** |
| Phone memory | 503 → 755 MB in 90 s | **~360 MB, flat** |
| Inbound frames, steady state | ~190/s | **low single digits** |

The gate's load is now that target and it passes: `docs/perf/0.1.25-target-100-30.json`.

## Two harness bugs the harness found

- Maestro ran under `spawnSync`, freezing the process that also served the fake Herdr and ran the sampler: every host call stalled 15 s, `terminal.attach` returned `unavailable`, and the phone showed **reconnecting**. It looked like an app defect and was not. Same requests now take 10-13 ms.
- The fake churned `agent.title` — the agent's **task** title — at 2 Hz. Real Herdr animates the *terminal* title. Because `taskTitle` is a stable-signature field, the fixture bypassed the host's coalescer and manufactured part of the flood it was measuring.

## Try it

[pr-205-caca5c62-vc56](https://github.com/umeranjum17/muxr/releases/tag/pr-205-caca5c62-vc56) — **arm64-v8a** for phones, x86_64 for the emulator.


umeranjum17:
## ARM64 recovery test APK

[Download the corrected APK](https://github.com/umeranjum17/muxr/releases/download/pr-205-aa92b014-arm64/app-release.apk)

> Corrected after install failure: the first APK used `versionCode 48`, but the current Android release is `versionCode 53`, so Android rejected it as a downgrade. This replacement is `versionCode 54` and uses the same signing certificate.

- Source commit: `aa92b014`
- Package: `com.trymuxr.app`
- Version: `0.1.25` (`versionCode 54`)
- ABI: `arm64-v8a` only
- Signing certificate SHA-256: `9c33841142483611257fcdf772f5e8fc30d6fac803f6da28e7eacfcaec12b08d`
- APK SHA-256: `f5894cc5a2bc26e6217cc23a4e8d771ef7ddff0b852fc1822c10ac1459153fb2`
- Durable GitHub release: https://github.com/umeranjum17/muxr/releases/tag/pr-205-aa92b014-arm64

### What to test

1. **Install over the current muxr app.** This APK has a higher version code and the same signing certificate, so it should upgrade in place and preserve pairing. Do not uninstall first.
2. **Normal terminal flow:** open muxr, wait for Connected, open and close several terminals, and send a harmless prompt in each. No terminal should remain on Reconnecting.
3. **Background recovery:** leave muxr in the background for 30–60 seconds, return, open a terminal, and send another harmless prompt. It should work without restarting the app or host.
4. **Host restart recovery:** keep muxr open on the phone, then run `systemctl --user restart muxr.service` on the host. Wait for the app to reconnect, then open a terminal and send a prompt. It should recover without restarting the app or workstation.
5. **Timeout privacy:** if a request catches the outage, the expected message is `request timed out: session.prompt (connection reset; try again after muxr reconnects)`. It must not expose a `machine-*` value, pane/session IDs, `MUXR_MACHINE_ID`, URLs, tokens, or raw socket-close text.
6. **Connection diagnostics:** during or immediately after the outage, open Settings → Connection. The latest bounded failure stage/code should be useful and sanitized; no internal IDs or secrets.
7. **Graphics recovery:** if a direct graphics surface stops and the `Graphics stopped` banner appears, tap Retry. Graphics should resume and the banner should clear after the host reports graphics ready; desktop resizing should continue to track the phone surface.
8. **Attachment cache:** open the same image attachment twice, including the full-screen/gallery view. The second open should render normally with no partial or corrupt image.

If recovery gets stuck, report the exact time, the last visible connection state, and a screenshot of Settings → Connection.

umeranjum17:
## Test builds + emulator profiling for `aa92b014`

Downloads (same commit as the previous asset — no fix included yet):

- Phone: [`muxr-0.1.25-55-arm64-v8a.apk`](https://github.com/umeranjum17/muxr/releases/download/pr-205-aa92b014-vc55/muxr-0.1.25-55-arm64-v8a.apk) · `sha256 3a1a5c80cbfa6c26286b158271419d1785f52eb475f332b2e036e4aeef0726c1`
- Emulator: [`muxr-0.1.25-55-x86_64.apk`](https://github.com/umeranjum17/muxr/releases/download/pr-205-aa92b014-vc55/muxr-0.1.25-55-x86_64.apk) · `sha256 745348d2d8213b2d39e496a9e17c6c056fb30de50c13122a732ba5d04e4bb6fb`

`0.1.25` / `versionCode 55`, package `com.trymuxr.app`. Signed with the local internal-testing key (cert SHA-256 `f1d45ed68e16ff6c93c8c5de8b20682e62724afc46a96df879a7b19bbc4afe57`), which is **not** the key behind `pr-205-aa92b014-arm64` (`9c338411…`). Uninstall the current app first; an in-place update is rejected as a signature mismatch.

## Reproduced on an emulator

Harness: headless `muxr_sandbox` (API 36, x86_64), permissive relay on `:8811`, isolated host in `MUXR_MODE=local`, Metro dev client, `adb reverse tcp:8081 tcp:8811`.

Deep-linking a bare shell reproduces the reported screenshot exactly: **Shell / Terminal** header, black viewport, `connecting` pill, forever. Host diagnostics record `client.hello`, `machines.list`, `session.list` — and **never `terminal.attach`**. `libexpolibghostty.so` loads fine, so the Kitty renderer is not implicated.

## Root symptom: the JS thread is saturated after authentication

| Measurement | Login screen | Authenticated |
|---|---|---|
| `mqt_v_js` CPU | idle (main 4%) | **96–104% continuously** |
| Native heap | flat | **+1.6 MB/s** (280 → 356 MB in 60 s) |
| TOTAL PSS | ~560 MB | 599 → 690 MB in 60 s, peaks observed at 1.1 GB |
| `gfxinfo` | — | 2 frames rendered, 0 janky, p99 7 ms |

No GPU or UI-thread jank: the UI is starved, not slow to draw. On a phone this ends in an LMK kill, matching “nothing loads at all”. Also thrown roughly once per 1.5 s:

```
Uncaught (in promise) Error: Maximum update depth exceeded
```

CDP pause-on-exception captured the enqueue path:

```
getRootForUpdatedFiber ← enqueueConcurrentRenderForLane ← forceStoreRerender
  ← zustand setState ← applySessions ← handleSessionEvent (async)
```

## What the loop is not

Each ruled out by measurement, not inspection:

- **Not a React re-render storm** — `react/jsx-runtime` `jsx`/`jsxs` instrumentation never reached 20 000 calls in 40 s.
- **Not `useState`/`useReducer`** — wrapped dispatchers captured no throw.
- **Not a store-write flood** — a side-effect-only breakpoint on zustand `setState` recorded ~2.7 writes/s, all `applySessions ← handleSessionEvent`.
- **Not inbound relay traffic** — instrumented `WebSocket` saw <200 messages in 40 s.
- **Not effect churn** — wrapped `useEffect`/`useLayoutEffect` never exceeded 20 runs per 3 s window.
- **Not screen-specific** — 104% CPU persists on Settings, and persists with the live-terminals strip rendered outside the `SectionList`.

## Where the time actually goes

`Debugger.pause` and the Hermes sampling profiler are both unusable once the thread saturates (the inspector target disappears; `HermesInternal.enableSamplingProfiler` is absent in this build), so native stacks were sampled with `debuggerd -b` on `mqt_v_js` (12 samples):

- 9/12 inside `libhermesvm.so` — JS executing, reached via `RuntimeScheduler_Modern::performMicrotaskCheckpoint` → `runEventLoopTick`, i.e. a **runaway microtask chain**, which is why timers, renders, and the inspector all starve.
- 2/12 inside `libunistyles.so` — `margelo::nitro::unistyles::HybridShadowRegistry::link(...)` allocating `UnistyleData` (`react-native-unistyles` 3.1.1).
- 1/12 `libjsi.so`.

`cpu-cycles`/`cpu-clock` perf events are unavailable in this emulator, so simpleperf could not be used.

## Secondary findings

1. The published `app-release.apk` cannot start on an x86_64 emulator: `SoLoaderDSONotFoundError: couldn't find DSO to load: libreactnative.so` from `MainApplication.onCreate` — the APK ships only `lib/arm64` while `DirectApkSoSource` resolves `lib/x86_64`. Hence the x86_64 asset above.
2. `local` mode can never attach a terminal: `relayTicket()` returns `undefined` for an empty token, so `openTerminal` throws `terminal: relay ticket required` before sending `terminal.attach`.
3. `useActivityAcknowledgements.markSeen` allocates a fresh `Set` even when no id changes, re-triggering renders and `AsyncStorage` writes.

## Next step

Pin the microtask chain (candidates: promise loops in the sync layer and the unistyles shadow-registry link path), then verify with this harness that `mqt_v_js` drops to idle and native heap stays flat. `react-native-unistyles` 3.2.0 fixes a shadow-registry cleanup leak and is worth testing as an upgrade, though the published fix is described as web-only.

umeranjum17:
## Load test on a release build, real machine, paired emulator

Two fixes are now on the branch: `ec1c04ae` (host coalesces title-only
`session.updated`, phone keeps `createdAt` and reuses unchanged card objects)
and `9dfcb44f` (phone merges inbound session frames into one store write per
250 ms).

**Load**: 20 freshly started `pi` agents, 8 panes animating their titles at
10/s, one pane streaming Kitty graphics through the host bridge, ~30
`session.updated`/s inbound measured in-app.
**Flow** (6 passes, Maestro): attach an agent's terminal, drag its scrollback
top to bottom 10x fast, detach, walk Machine/Usage/Files/Panes, open the
graphics pane, scroll the strip and the spaces tree, leave the app, come back.

| Build | Flow | JS thread | Frames | Update-depth errors | Screen |
|---|---|---|---|---|---|
| `ec1c04ae` (no batching) | fails on pass 1 - herd never appears | never sampled: runtime down, 7 restarts | **0/s over 7 min** | 0 logged before the runtime died (5 in an earlier run) | blank white |
| `9dfcb44f` (batching) | **all 6 passes complete** | **31.2%** over 380 s sampled | 23.6/s | **0** | renders throughout, 20 agents listed |

Steady state, same builds, no navigation: `9dfcb44f` sits at 19.4% JS / 58 fps
for two minutes at 30 frames/s inbound; `ec1c04ae` dies inside 20 s.

### Where the cost was

Measured in-app, not guessed: offline the JS thread costs 7%, online 43%, and
38% with the app backgrounded and **zero frames drawn** - so it was never
rendering. Per inbound frame: decode ~4 ms, merge ~0 ms, `deepEqual` ~0 ms,
list rebuild ~0.4 ms. The expense was one React pass per frame, 12-30 times a
second. Batching removes the passes, not the work.

Note on tooling: `adb shell top -H -n 1` reports a thread's *lifetime* average,
so a saturating build and a healthy one measured identically (96% vs a real
63%). All numbers here come from `/proc/<pid>/task/<tid>/stat` jiffie deltas.

### Still broken, unchanged by these commits

Kitty graphics never paint on the phone. With `terminal-browser` rendering
Example Domain in a pane, the attached terminal shows the text stream and an
empty viewport where the image belongs - the same symptom the earlier phase-3
QA recorded at `434dfb3f` ("phone Ghostty surface stayed solid black"). Not a
regression from this branch, and not fixed here.

umeranjum17:
## Kitty graphics now paint on the phone

The blank image area was not a rendering bug on the phone and not a regression
from this branch. The host's graphics bridge only handled Herdr's leased-file
transport (`ServerMessage::GraphicsFile`, type 13). Captured from the live
client socket while a terminal browser painted in a pane:

```
msg=welcome version=20
msg=other code=2 len=2904309   ...\e_Ga=t,t=d,f=32,s=616,v=875,i=801812,q=2,m=1;...
code=13 messages: 0
```

Herdr 0.8.2 delivers a program's own Kitty images **inline in the app output
stream**, positioned by the cursor. The bridge dropped that stream by design,
so the phone got text over an empty viewport - the same symptom the phase-3 QA
recorded at `434dfb3f` as "surface stayed solid black".

`6bf683f6` adds `inlineKitty.ts`: it follows absolute cursor moves, joins
chunked transmissions, and returns each APC block with the cell it landed on.
The bridge routes placements through the existing `pane.layout` query and feeds
the decoded pixels into the same `prepareKitty` / `encodeKitty` path the file
transport used, so the phone gets one compressed, pane-fitted frame.

Three things the emulator taught us, all in the commit:
- image data is learned even with no phone attached (programs transmit first)
- unchanged placements are dropped (a repainting producer emits ~30/s)
- frames must be compressed and self-contained: forwarding the raw 2.9 MB RGBA
  tore the terminal socket into an attach loop, and the phone's write pump
  keeps only the newest graphics frame

**Evidence**: a browser rendering in a Herdr pane paints on the phone, and two
live repaints tracked (screenshots in the pane attachments). Host unit tests
8 files / 46 tests pass, including a new inline flow test.

## Reconciled load test

Everything at once: 20 freshly started agents, live Kitty graphics repainting,
6 panes animating titles, and 6 passes of Maestro driving an agent's terminal -
attach, drag the scrollback top to bottom 10x fast, detach, walk the plugin
tabs, open the graphics pane, leave the app, come back.

| | Result |
|---|---|
| JS thread | 26.0% over 290 s sampled |
| React update-depth errors | 0 |
| PSS | 267 MB, flat |
| JS runtime | alive, 6 app restarts all recovered |
| Agents | 20 alive and listed |
| Graphics | still painting after the run |

For contrast, the same flow on the pre-batching build (`ec1c04ae`) never
reaches the herd screen: the runtime dies inside 20 s, 0 frames rendered in
7 minutes, blank white screen.

## Notes

- The host fix only reaches your phone once the host ships: the running service
  uses the published CLI. Verified here by running this branch's host in place
  of the service, then restoring it.
- The Panes plugin header shows a stale pane count (33 vs 40 live) while the
  pane rows themselves are correct; `herdr.tree` itself is fresh - a new tab
  appears within 25 s. Cosmetic, not touched here.

umeranjum17:
## Release performance gate, on a herd we own

`yarn perf` now measures the release APK against **our real relay and our real host**, with Herdr faked at its three wire seams (`perf/fake-herdr`): the NDJSON control socket, the `HERDR_BIN` CLI, and the protocol-20 graphics socket. Relay, host and app are the builds we ship; only the third-party desk is simulated, so a release can be measured with the same load every run and without touching the machine's own herd.

The fake reproduces the behaviours the three shipped bugs lived in: per-pane `pane.updated` title churn that is visible in the next snapshot, `pane.read` previews, a terminal stream with a full repaint for every scroll and resize, and inline Kitty APC blocks chunked across frames.

### Gate result (30 panes, 6 agents, titles 2 Hz, graphics 4 Hz)

| Phase | JS thread | Frames | Memory |
|---|---|---|---|
| idle on the herd | 39.1% | 29.2 fps | 282 -> 302 MB |
| herd strip and tree soak | 41.8% | 41.2 fps | 299 -> 268 MB |
| agent terminal and plugin navigation | 31.4% | - | 280 -> 262 MB |

Plus, from the host's own journal: 100 requests, **56 terminal attaches, slowest request 13 ms, none rejected**, no React update-depth errors, runtime alive, 24/24 panes toured with 17 MB growth. Evidence: `docs/perf/0.1.25-vc55.json`.

### Two harness bugs the harness found

- Maestro ran with `spawnSync`, freezing the process that also served Herdr. Every host call then stalled 15 s, `terminal.attach` came back `unavailable`, and the phone showed **reconnecting**. The fake now runs in its own process and flow runs never block the sampler - the same requests are 13 ms.
- The sampler was frozen by the same call, so earlier fps and stall numbers measured the harness, not the app.

### Capacity (`node perf/capacitySweep.mjs`)

| Herd | JS thread | Host + relay RSS | Verdict |
|---|---|---|---|
| 10 panes / 2 agents | 13.3% | 246 MB | ok |
| 30 panes / 6 agents | 38.4% | 294 MB | ok |
| 45 panes / 9 agents | 56.2% | 303 MB | ok, at the edge |
| 60 panes / 12 agents | 75.9% | 325 MB | phone over budget |
| 90 panes / 18 agents | - | - | herd never finished reaching the phone |

The binding constraint is the phone's JS thread, not host RAM: host plus relay stay at 250-330 MB across the whole ramp (~0.4 MB per pane), so 2 GB is never the limit. Evidence: `docs/perf/capacity-0.1.25.json`.

One production change: `HERDR_SOCKET_PATH` in `apps/host/src/main.ts`, so a harness can point the host at its own Herdr instead of the desk's.

umeranjum17:
## 100 panes / 30 agents is now the gate's load, and it passes

Measured on the rig (real relay, real host, real release APK on the emulator, Herdr faked at its wire seams), same 100/30 herd before and after:

| | before | after |
|---|---|---|
| Phone JS thread | 91-97.6% | **36-43%** |
| Phone memory | 503 -> 755 MB in 90 s | **~360 MB, flat** |
| Steady-state inbound frames | ~190/s | **low single digits** |
| First sync (herd on screen) | 40-50 s | 36 s |
| Host + relay RSS | 330 MB | 330 MB |

Gate run at the target: `docs/perf/0.1.25-target-100-30.json` — 40/40 panes toured with 18 MB growth, 41 terminal attaches, slowest host request 10 ms, no rejections, no update-depth errors, runtime alive.

### What was actually costing it

Attribution came from a temporary per-second counter on the frame path, not from reading code. At 100/30 the app was receiving **190 frames/s** (`session.updated` 130/s, `lifecycle.update` 60/s) and spending a full second per second inside the handler.

1. **Every decrypted frame awaited a disk write.** `persistReplay` (`hostedE2ee.ts`) serialized the entire replay cache — one entry per session stream, so a hundred entries — and awaited `AsyncStorage` inside `open()`. The write chain could not drain, and each queued closure held its own copy: that was the memory climb. Now a trailing write (dirty flag, one in flight, flush on background), the pattern the relay already uses and that `stream` channels already had.
2. **The host published `terminalTitle` in every `session.updated`.** A working agent animates it several times a second; no client reads it (written at `sessionMapping.ts:81`, declared in the store schema, zero readers — card titles come from the agent's own title). Removed from `SessionInfo`, together with the per-session coalescing timers that existed only to throttle it. `herdr.tree` still carries it at the tree's cadence.
3. **`LifecycleStore.transition` republished the same lifecycle event** — same `eventId` — whenever only the task title moved, and `save()`d the journal each time. That was the 60 frames/s and 60 host disk writes/s. A title is not a lifecycle change.

Plus: the app pulled the whole herd tree one second after *any* session write; on a busy herd that is a large decrypt, parse and deep-compare for placement that rarely moves. Debounce is now 3 s.

### Fixture bug the same numbers exposed

`perf/fake-herdr` was churning `agent.title` — the agent's **task title** — at 2 Hz. Real Herdr animates the *terminal* title; a task title changes when the work does. Because `taskTitle` is a stable-signature field, the fake was bypassing the host's title coalescer entirely and manufacturing part of the flood. The fake now animates only the terminal title and changes task titles on a human timescale. The host fixes above stand on their own — they are what a real desk needs when an agent does update its title text.

### Known, not fixed

- Every 15 s the fake flips all 30 agents at once and the host answers with **five frames per agent** (`status.update`, `activity.update`, `lifecycle.update`, `attention.update`, `session.updated`) — a 150-frame burst that queues seconds of work. `attention.update` also carries the whole attention catalog per event. That is the next lever if bursts matter.
- `sessionSync.integration.spec.ts` "reconciles structured lifecycle activity" fails, and fails at `HEAD~3` too: inherited, unrelated to this change, left flagged rather than guessed at.

umeranjum17:
## Test build for the target load

APK from `caca5c62` (0.1.25, versionCode 56) - the commit that made 100 panes / 30 agents pass the gate:

**[muxr-0.1.25-vc56-arm64-v8a.apk](https://github.com/umeranjum17/muxr/releases/download/pr-205-caca5c62-vc56/muxr-0.1.25-vc56-arm64-v8a.apk)** - phone build, install this one.

The emulator build (x86_64) is on the [same release](https://github.com/umeranjum17/muxr/releases/tag/pr-205-caca5c62-vc56) and will not install on a phone. Signed with the usual release key, so it upgrades in place over the previous test build.

umeranjum17:

PR #200, opened 2026-09-01
feat: make phone connection setup automatic

umeranjum17:
## Summary

- replace the infrastructure-first connection menu with one detected, explained route proposal
- keep a healthy current route first, then rank Tailscale, existing private overlays, an installed temporary tunnel, and same Wi-Fi
- detect NetBird, WireGuard, ZeroTier, and generic private interfaces without adding provider-specific integrations
- keep every alternative under **Choose another way** and preserve the final **Apply setup** trust gate
- keep inconclusive Tailscale Serve checks advisory; only proven disabled or occupied Serve selects direct Tailscale
- update the public README, npm README, self-hosting guide, and bundled onboarding skill to match the shipped flow

This is stacked on #199 because it consumes the bounded Serve states introduced there. Merge #199 first; this PR can then be retargeted to `main`.

## User flow

1. muxr inspects the machine without changing it.
2. Setup shows one **Recommended route** with **Use this route and continue**.
3. **Choose another way** reveals private networks, same Wi-Fi, temporary tunnels, and custom servers.
4. The user reviews all changes and chooses **Apply setup**.
5. Setup verifies the route and shows the existing one-use pairing QR.

If Tailscale Serve becomes unavailable after Apply, setup offers direct Tailscale or stops with an accurate receipt; it no longer restarts the wizard or claims that nothing changed after prerequisites ran.

## Safety

- no network, service, integration, or plugin mutation occurs before **Apply setup**
- occupied Serve state is never overwritten
- existing healthy routes and pairings stay preferred
- Docker, VM host-only, and Tailscale duplicate interfaces are excluded from misleading LAN/private recommendations
- the new private-overlay mode uses the selected private address for the QR; pairing and E2EE remain authoritative

## Verification

- `mise x node@22.23.2 -- yarn build`
- `node scripts/diagnostics/application/checkTailscaleIngress.mjs`
- `node scripts/diagnostics/application/checkArchitecture.mjs`
- `node scripts/diagnostics/application/checkPackageSmoke.mjs`
- temporary-state flow QA: private relay reached QR generation; invalid ports created no state or mutation commands
- read-only review approved after the post-Apply receipt, shared-relay preflight, macOS duplicate-route, VM adapter, and documentation findings were repaired

## Deliberate non-goals

- SSH-carried pairing is separate contract/mobile work and is not added here.
- Exact interface-only relay binding is not changed here; direct private/LAN modes retain the existing listener behavior, with muxr pairing and E2EE still required.
- No provider is installed or configured automatically.

PR #196, opened 2026-08-31
Cut bundled plugins 14→8 and drive pane layout through Herdr actions

umeranjum17:
## Current head

`c23216a5f192481063b1b37b83b47cee17dffc82` — `fix(plugins): drop unused Runbook and broken Preview product UI`.

Removes the unused Runbook surface and the phone Browser Preview / Ports product paths that never worked on a remote pair (`preview: relay ticket required`). The graphical takeover / AppDirectGraphics tunnel stays: `preview.attach`, relay `/preview`, crypto seal/open, and `attachPreviewTunnel`. Code is now read-only. Setup prunes a leftover `runbook.mjs` on upgrade. `kernel.navigate` preview and terminal port chips are gone.

Local `yarn run check`: **32/34**. Only the documented Agent-close baseline failures remain. Takeover-tunnel diagnostic: **PASS**. Fresh x86_64 debug APK SHA-256 `40e80ddda3f758392b05721595aea07f5daedd6dda3c5fd69f7072c667ddfb0a`. Proof files live in the worker pane attachments (`PR196-PROOF-MANIFEST.md`, `PR196-09`–`15`). No paired plugin-nav recapture: that catalog is live Herdr, and linking this worktree would mutate the owner registry.

Fable/Sol reviews are not part of this push. Do not merge.

---

Cuts the bundled plugin set from 14 to 8, makes muxr's first real use of the Herdr action contract, and fixes the ccusage flakiness.

The full audit that motivated this is owned by another agent (`docs/plugin-architecture-review.md`) and is deliberately not in this PR.

## What changed and why

**14 → 8 bundled plugins.** Deleted `example-ui` (an SDK sample that shipped *enabled*, putting a dummy screen in the real app), `servers` (Preview is already reachable from the terminal port chip, later removed on this head) and `inbox` (Recent Activity renders the same done/blocked/waiting states). Folded `voice-gemini`/`voice-openai`/`voice-codex` into `plugins/voice/providers/*.mjs`.

**Deleted the `RETIRED` map.** Replaced by a rule with no table: a registration pointing into our bundle directory that no longer names a shipped plugin is unlinked on setup. Anything the user linked from elsewhere is untouched.

**Provider selection moved into the plugin.** Removed the `voice.provider.list`/`voice.provider.select` host requests. The phone calls the plugin's capability through `callPlugin`. View-only devices are refused write-mode RPCs.

[...]

PR #195, opened 2026-08-31
feat: add generic Android Herdr graphics and panes

umeranjum17:
## What ships

- `plugins/panes`: generic **Tools** and **Panes** surfaces for terminal programs, agent-less panes, enabled `[[panes]]` contributions, and actions that explicitly declare `contexts = ["global"]`.
- Android-only generic Herdr 0.8.2 `AppDirectGraphics` bridge for pane-scoped `GraphicsFile` producers.
- Standard Kitty bytes into the existing Android libghostty terminal, with physical cell geometry, aspect-fit placement, letterbox-aware pointer mapping, tap/drag/pinch arbitration, and authoritative graphics retirement.
- Serialized conversion plus latest-frame coalescing on the Herdr leg; 512 KiB/128 KiB high/low watermarks and one pending latest graphics frame on the phone WebSocket leg.
- Ordinary ANSI/control remains on `herdr terminal session control`; iOS and web remain ANSI-only.

No producer-specific transport exists. The same path is runtime-proven with two independent upstream tools: `zenbu-labs/terminal-browser` and `zenbu-labs.tode`/terminal-code.

## Runtime acceptance

Fresh rebuilt Android app, cleared app data, isolated PR host/relay, Herdr 0.8.2 protocol 20:

- terminal-browser: real Chromium render, tap navigation, keyboard activation, drag-scroll.
- terminal-code: full VS Code render and touch-opened README.
- concurrent panes: no image crossover.
- producer exit: Kitty placement retired and the same pane recovered to text.
- nvim: touch/IME, normal/insert mode, scroll, key row, and `:w` regression passed.
- pinch: controlled grid changed from 56×42 to 54×39.

## Measured performance

- Host direct-frame delivery: **59–61 fps**; normal-link WebSocket `bufferedAmount`: **0**.
- Android display recording: **28 frames / 0.989 s = 28.3 fps**.
- Host: **45.7% of one CPU core**, **224,276 KiB max RSS**.
- At 1 Mbps, buffering remained bounded and latest-frame coalescing replaced about **60 frames/s**; host RSS remained flat.

## Out of scope

Herdr and its worktrees are strictly read-only. This PR supports pane-scoped Herdr `GraphicsFile` producers only. Arbitrary PTY Kitty capture (including mpv/yazi images), protocol-level ≤2 s recovery guarantees at 1 Mbps, partial/multi-image layouts, and iOS/web graphics are not supported or claimed. The proof manifest records those boundaries.

## Reviewer dispositions

[...]

PR #142, opened 2026-08-27
feat(store): add approved Apple screenshot sets

umeranjum17:
## Problem

The App Store listing had only onboarding placeholders. Release screenshots must show the strongest current product flows on genuine iPhone and iPad layouts without stale versions, private data, internal route labels, or transient errors.

## Changes

- add four Fable-approved iPhone 6.9-inch compositions: Herd, terminal, files, Runbook
- add three Fable-approved genuine iPad 13-inch compositions: Herd, files, Runbook
- preserve every authentic exact-main raw capture used by the compositions
- optimize the canonical keyword field to 99 characters without duplicating title/subtitle terms

## Evidence

- source app commit: current main `bccd0e0a`
- iPhone outputs: 1320×2868, opaque, 8-bit sRGB
- iPad outputs: 2064×2752, opaque, 8-bit sRGB
- UI is unaltered and uncropped; only restrained outer framing and headlines are added
- Fable review: approved after two rounds
- metadata validation: 0 errors, 0 warnings
- secret scan: pass

Rejected captures with internal route labels, action/reconnect overlays, or blank/disconnected terminals are intentionally excluded.

PR #209, opened 2026-09-03
release: 0.1.26 consolidated fixes and local validation

umeranjum17:
Release 0.1.26 makes the phone a more reliable window into work on the computer: connect, find live work, read it, understand usage, and direct it by voice.

**Merged as `740e77193940a67b9875afb8eb2403b12d482273`. The merge tree matches tested candidate `d40890f049eae61ffd9b340af64a403349ae7175`; all candidate GitHub checks and the fresh full local emulator gate PASS.** The shared voice watchdog now has a fixed ceiling; the live probe retains lock/scratch when browser absence is unproven. ARM64 phone and real speech/browser acceptance remain distinct from the passing x86 emulator run.

This guide uses a small diagram for each feature or reliability change. Expand a card to see **why it exists → what changes → what was actually tested**. Diagrams describe implemented behavior, not proof that every branch has been exercised live.

### System view

```mermaid
flowchart TD
 U[User reads navigates and speaks] --> P[Phone UI and native audio]
 P <-->|Authenticated work connection| H[muxr host]
 H <--> R[Herdr live work and plugins]
 R --> W[Agents files terminal output and graphics]
 W --> H
 P <--> V[Selected voice provider adapter]
 V --> K[Shared voice tool runtime]
 K --> H
 K --> P
```

### How to read the evidence

| Label | Meaning |
| --- | --- |
| **LOCAL** | Real source/module/package checks. Specific focused counts may be pinned to component commits; the latest repository suite passes at `bc90a182`. They are not device or live-provider proof. |
| **EMU32** | Full local emulator PASS at `32edfc54`, with matching native/APK/source evidence and controlled topology/usage fixtures. New assistant mobile work landed later. |
| **LIVE-PARTIAL** | Earlier real device/provider observations; only the behavior explicitly named is established. |
| **DEPLOYED** | Local backend package `bc90a182` installed and service/plugin identity checked. Deployment does not establish a completed voice exchange. |
| **SOURCE / OPEN** | Included implementation or materials / outstanding acceptance or correctness gap. |

### Feature map

- **Connect to the computer:** 1. Choose a usable connection route; 2. Recover from unavailable Tailscale Serve; 3. Diagnose installation and service drift; 4. Recover a connection whose first host frame never arrives.
- **Discover and run capabilities:** 5. Generic Tools and native plugin screens; 6. Understand and manage plugin roles; 7. Use Panes actions and see launch failures; 8. Grant private Usage context through the installed registry.
- **Read files and changes:** 9. Navigate repository files, folders and paths; 10. Read code and diffs on one surface; 11. Pan long lines and wrap Unicode correctly; 12. Zoom and keep controls reachable on narrow screens; 13. Navigate Git files, hunks and word changes; 14. Show truthful source labels while scrubbing; 15. Open the requested source line; 16. Recover cold and offline relative file links.
- **Render terminals and graphics:** 17. Render graphics from Herdr and the terminal PTY; 18. Preserve placements and deletes under backpressure; 19. Retire graphics without resurrecting stale images; 20. Keep input, zoom and image geometry aligned.
- **Keep the herd responsive:** 21. Batch work updates without publishing false lifecycle changes; 22. Scroll without a React update for every frame.
- **Understand provider usage:** 23. Read OMP and OpenCode activity across profiles; 24. Order tabs by recent activity in the correct date window; 25. Distinguish local tokens from Go quota availability; 26. Keep cached usage scoped to the account.
- **Configure realtime voice:** 27. Choose one provider and preserve that choice; 28. Open the setup screen belonging to the selected provider; 29. Keep realtime audio native and service-ordered; 30. Choose only a currently live voice target.
- **Direct work through the assistant:** 31. Inspect useful work metadata before asking for a name; 32. Find agents with conservative fuzzy names and task descriptions; 33. Use phone focus, recent views and semantic navigation; 34. Send complete prompts and report delivery honestly; 35. Read real output to summarize ongoing work; 36. Share tool execution and completion behavior across providers; 37. Turn acknowledgement into a result or explicit failure.
- **Prove behavior locally:** 38. Build an APK with source and native provenance; 39. Exercise actual feature mounting end to end; 40. Measure performance with bounded, inspectable evidence; 41. Separate real voice/browser probes from fixture acceptance; 42. Preserve failures and keep fixtures deterministic.
- **Ship an understandable package:** 43. Package and deploy the consolidated release; 44. Explain voice data use and supply store assets; 45. Remove obsolete surfaces and retain scoped maintenance updates.

### Connect to the computer

<details>
<summary><strong>1. Choose a usable connection route</strong></summary>

**Why:** A phone needs a reachable route to the computer before any work can be useful.

**Behavior:** Setup probes Herdr, agent integrations and network options, then proposes ready private-network, Tailscale or LAN routes, with tunnel/external alternatives. Installed software and a functioning service are separate findings.

```mermaid
flowchart LR
 A[Inspect machine and network] --> B[Classify available routes]
 B --> C[Propose ready routes]
 C --> D[User chooses connection]
 D --> E[Apply setup]
```

**Tested / limits:** LOCAL: setup and connection flows are in the passing repository suite. EMU32 proves its controlled connection path; it does not exercise every real network configuration.

</details>

<details>
<summary><strong>2. Recover from unavailable Tailscale Serve</strong></summary>

**Why:** An unavailable convenience route should explain the next step without damaging another service.

**Behavior:** Disabled Serve produces an enable/alternative path. Inconclusive checks stay advisory, and an already occupied route is preserved.

```mermaid
flowchart TD
 A[Check Serve] --> B{Result}
 B -->|Ready| C[Offer Serve]
 B -->|Disabled| D[Explain enable or alternative]
 B -->|Inconclusive| E[Keep advisory]
 B -->|Occupied| F[Preserve existing route]
```

**Tested / limits:** LOCAL: focused setup checks and final repository suite pass. No claim that every Tailscale policy or remote machine was tested live.

</details>

<details>
<summary><strong>3. Diagnose installation and service drift</strong></summary>

**Why:** The CLI, running service and registered plugins must describe the same installation.

**Behavior:** Setup/doctor distinguishes CLI and service versions, checks Herdr readiness and repairs missing bundled registrations through normal setup. Bundled registration is not silently performed by live probes.

```mermaid
flowchart LR
 A[CLI package] --> D[Compare installation state]
 B[Running service] --> D
 C[Registered plugin roots] --> D
 D --> E[Report drift or apply normal setup]
```

**Tested / limits:** LOCAL: build, package smoke and bundled guards pass. DEPLOYED: the local host, relay and nine bundled roots were checked against the bc90a182 package after restart.

</details>

<details>
<summary><strong>4. Recover a connection whose first host frame never arrives</strong></summary>

**Why:** An open WebSocket is insufficient evidence of an authenticated working connection.

**Behavior:** The initial liveness deadline now retires and closes an unproductive socket, letting normal reconnect run. Socket identity and post-await guards reject late work from old connections; crypto and permanent-refusal rules remain enforced.

```mermaid
flowchart TD
 A[Socket opens] --> B{Authenticated host frame}
 B -->|Arrives| C[Connected]
 B -->|Deadline expires| D[Retire and close socket]
 D --> E[Normal reconnect]
 E --> B
```

**Tested / limits:** LOCAL: existing liveness flow 4/4 and generated-key encrypted recovery passed; baseline failure was reproduced. Earlier device recovery occurred without manual reconnect. EMU32 includes the corrected client, but this is not a universal network reliability claim.

</details>

### Discover and run capabilities

<details>
<summary><strong>5. Generic Tools and native plugin screens</strong></summary>

**Why:** A new extension should expose capabilities through a shared contract instead of requiring its own phone implementation.

**Behavior:** The phone discovers plugin contributions and renders supported native primitives. Herdr owns executable registration; backend plugins own behavior. Declarative screens do not download arbitrary React code.

```mermaid
flowchart LR
 A[Installed plugin manifest] --> B[Contribution catalog]
 B --> C[Generic Tools and native primitives]
 C --> D[User action]
 D --> E[Host and Herdr plugin runtime]
```

**Tested / limits:** LOCAL: bundled catalog, primitive and architecture guards pass. EMU32 mounts real Files/Usage implementations through a controlled catalog; arbitrary third-party plugins are not all validated.

</details>

<details>
<summary><strong>6. Understand and manage plugin roles</strong></summary>

**Why:** Users need to distinguish an executable capability from a phone screen and know whether it is enabled.

**Behavior:** Settings separates backend-only, UI-only and combined contributions. Bundled registration/pruning is consolidated, saved voice selection is preserved, and revocation is honored.

```mermaid
flowchart TD
 A[Plugin contribution] --> B{Declared role}
 B --> C[Backend capability]
 B --> D[Phone UI]
 B --> E[Both]
 C --> F[Enabled and revoked state]
 D --> F
 E --> F
```

**Tested / limits:** LOCAL: plugin/package and voice lifecycle guards pass, including packaged configuration flows. The live catalog must be queried by full plugin ID; a partial-name filter is not absence evidence.

</details>

<details>
<summary><strong>7. Use Panes actions and see launch failures</strong></summary>

**Why:** An action is useful only if its actual result reaches the user.

**Behavior:** Panes discovery and declarative split actions use the shared plugin path. Short-lived action failures are surfaced instead of treating process launch alone as success. Files receives topology through declared host context.

```mermaid
flowchart LR
 A[Choose pane action] --> B[Resolve declared context]
 B --> C[Execute through Herdr]
 C --> D{Observed outcome}
 D --> E[Started or completed]
 D --> F[Actionable failure]
```

**Tested / limits:** LOCAL: plugin action/context checks and the repository suite pass. This is not a claim that every third-party action was tapped on the final phone build.

</details>

<details>
<summary><strong>8. Grant private Usage context through the installed registry</strong></summary>

**Why:** A plugin may need a narrow private input without inheriting the host environment or choosing its own privileges.

**Behavior:** A module-anchored package registry matches canonical entry, plugin ID and method. Entries must remain inside the package plugins root. Trusted recipes overwrite reserved stdin context; subprocess environment remains closed. Go receives only the supported type/key projection, bounded to 16 KiB.

```mermaid
flowchart TD
 A[Plugin invocation] --> B{Installed canonical registry match}
 B -->|Yes| C[Validate narrow recipe]
 C --> D[Replace reserved stdin context]
 B -->|No| E[No private grant]
 D --> F[Child with restricted environment]
 E --> F
```

**Tested / limits:** LOCAL: policy/catalog checks, forged-entry cases and account/config flows passed. EMU32 exercises the real private-context implementation through isolated fixtures. No live Go quota success is fabricated.

</details>

### Read files and changes

<details>
<summary><strong>9. Navigate repository files, folders and paths</strong></summary>

**Why:** A file link must resolve in the intended repository, and a folder should remain a navigation destination.

**Behavior:** Breadcrumbs, file navigation, deep paths, quoted Unicode names, copy chips and third-party diff context are corrected. Directory results navigate as directories rather than surfacing a file-read errno.

```mermaid
flowchart TD
 A[File or folder link] --> B[Resolve repository context]
 B --> C{Destination}
 C -->|Directory| D[Browse folder]
 C -->|File| E[Open document]
 E --> F[Breadcrumb and adjacent file navigation]
```

**Tested / limits:** LOCAL: existing mobile file/diff flows and checks pass. EMU32 proves a real session-relative file route and native diff opening; every path/copy variant has not been individually demonstrated on device.

</details>

<details>
<summary><strong>10. Read code and diffs on one surface</strong></summary>

**Why:** The same document should retain one coherent reading area as its mode changes.

**Behavior:** Code, document and diff views share the reading surface, header and navigation. The edge-to-edge column and code palette improve available width; wrapped rows no longer invent a continuation arrow.

```mermaid
flowchart LR
 A[Document content] --> C[Shared reading surface]
 B[Git changes] --> C
 C --> D[File mode]
 C --> E[Diff mode]
 D --> F[One header and reading column]
 E --> F
```

**Tested / limits:** LOCAL: existing viewer flows and touched-file type checks passed. EMU32 opens a real Git diff through Session actions and switches to File; native document scrolling is measured.

</details>

<details>
<summary><strong>11. Pan long lines and wrap Unicode correctly</strong></summary>

**Why:** Source text width is determined by display cells, not the number of UTF-16 code units.

**Behavior:** Wrap can be disabled for horizontal panning. Expanded tabs and display-cell width size the surface, so the tested 80-character CJK line remains one unwrapped native Text row.

```mermaid
flowchart LR
 A[Source text] --> B[Expand tabs and measure cells]
 B --> C{Wrap selected}
 C -->|Yes| D[Visual wrapped rows]
 C -->|No| E[One wide row and horizontal pan]
```

**Tested / limits:** LOCAL: existing code-layout flow covers the CJK regression. EMU32 requires CJK start/end in the same Text node and ASCII-equivalent height, plus visible horizontal movement. Full horizontal extent remains device-unverified.

</details>

<details>
<summary><strong>12. Zoom and keep controls reachable on narrow screens</strong></summary>

**Why:** Larger text should preserve navigation and touch targets rather than push controls off screen.

**Behavior:** Zoom follows a ladder. File/hunk counters give way on narrower rows; below the compact breakpoint the zoom pair moves into the overflow menu while keeping its disabled state and full touch targets.

```mermaid
flowchart TD
 A[Available screen width] --> B{Control budget}
 B -->|Wide| C[Navigation counters and zoom]
 B -->|Narrow| D[Drop visible counters]
 B -->|Compact| E[Put zoom in overflow menu]
 C --> F[Reachable controls]
 D --> F
 E --> F
```

**Tested / limits:** LOCAL: compact-width arithmetic and implementation review passed. EMU32 checks simultaneous file/hunk controls for overlap and verifies increased line height after zoom. Every supported width is not device-proven.

</details>

<details>
<summary><strong>13. Navigate Git files, hunks and word changes</strong></summary>

**Why:** A diff must retain the relationship between changed files, changed regions and their source lines.

**Behavior:** File/hunk navigation, folded context and inlay rows share the document surface. Word marks are limited to equal add/delete counts; empty diffs and deep paths remain readable.

```mermaid
flowchart LR
 A[Real Git baseline and changes] --> B[Files]
 B --> C[Hunks and context]
 C --> D[Old and new source rows]
 D --> E[Word marks where pairing is valid]
```

**Tested / limits:** LOCAL: existing inlay/layout flows pass with a committed fixture. EMU32 requires removed and added markers from an actual Git diff and simultaneous file/hunk navigation. The equal-count word-mark gate lacks dedicated visual proof.

</details>

<details>
<summary><strong>14. Show truthful source labels while scrubbing</strong></summary>

**Why:** A scrollbar label should identify the row the list actually reaches, even across folds and patch gaps.

**Behavior:** The scrubber and list use the same scrollable-pixel target. Padding is removed before binary-searching row offsets; labels use real new/old source lines and skip gap/pill items.

```mermaid
flowchart LR
 A[Drag fraction] --> B[Actual scrollable pixel offset]
 B --> C[Subtract top padding]
 C --> D[Find real row by offset]
 D --> E[Show carried source line]
```

**Tested / limits:** LOCAL: mapping arithmetic and focused code review passed. Exact source labels under device scrubbing remain unverified; this is not covered merely by a scrolling performance PASS.

</details>

<details>
<summary><strong>15. Open the requested source line</strong></summary>

**Why:** Line 200 means source line 200, not visual row 200 after deletions, wraps or folds.

**Behavior:** The route forwards line=N to the surface. Lookup prefers new-line numbering and uses old-line fallback only for deleted rows. A folded target expands; a real jump waits for positive viewport layout and keeps context above the target.

```mermaid
flowchart TD
 A[Link with source line] --> B[Find actual numbered row]
 B --> C{Inside folded context}
 C -->|Yes| D[Expand and rebuild rows]
 D --> E[Wait for viewport layout]
 C -->|No| E
 E --> F[Scroll with preceding context]
```

**Tested / limits:** LOCAL: inlay flow covers source-index drift and the deleted-line collision. EMU32 proves cold and warm relative line-200 placement. Folded/deletion-collision variants are not all device-proven.

</details>

<details>
<summary><strong>16. Recover cold and offline relative file links</strong></summary>

**Why:** An unresolved session-relative path must never be guessed against the host working directory.

**Behavior:** The route subscribes to session metadata, waits for its root, and shows an in-screen unavailable error after an eight-second target-keyed wait or a completed missing-root catalog. Retargeting resets the wait; late hydration recovers. Absolute paths bypass the root wait.

```mermaid
flowchart TD
 A[Relative file link] --> B{Session root available}
 B -->|Yes| C[Resolve absolute path and load]
 B -->|No| D[Subscribe and wait for this target]
 D -->|Root arrives| C
 D -->|Timeout or missing session| E[In-screen unavailable]
 E -->|Later hydration| C
```

**Tested / limits:** LOCAL: reactive-state/timeout transitions were reviewed. EMU32 proves cold/warm relative line 200, bounded offline missing-root error, then restored connection and line-200 recovery. The fixture remains relative to exercise session resolution.

</details>

### Render terminals and graphics

<details>
<summary><strong>17. Render graphics from Herdr and the terminal PTY</strong></summary>

**Why:** Images emitted by terminal programs need a real route to the phone, not only browser-specific metadata.

**Behavior:** The release contains generic GraphicsFile/AppDirectGraphics handling and inline PTY Kitty decoding, with Android/iOS/web seams. Synthetic graphics and real browser pages are separate acceptance targets.

```mermaid
flowchart LR
 A[Herdr graphics files] --> C[Host graphics bridge]
 B[PTY Kitty output] --> C
 C --> D[Pane-bound transport]
 D --> E[Mobile or web renderer]
```

**Tested / limits:** LOCAL: host/mobile graphics flows and native patch checks pass. EMU32 proves native checkerboard pixels and delivered frames. Real GitHub body paint, comprehensive iOS/web and physical-phone behavior remain open.

</details>

<details>
<summary><strong>18. Preserve placements and deletes under backpressure</strong></summary>

**Why:** A delete and two different placements are ordered operations; replacing them with only the latest message loses meaning.

**Behavior:** An ordered FIFO replaces transport-level latest-only dropping. The channel has 64 MiB/128-frame hard limits and closes explicitly on overflow. Payload/frame handling also has separate size bounds.

```mermaid
flowchart LR
 A[Placement A] --> Q[Ordered bounded queue]
 B[Delete A] --> Q
 C[Placement B] --> Q
 Q --> D[Drain in order]
 Q -->|Hard bound exceeded| E[Explicit channel close]
```

**Tested / limits:** LOCAL: exact-source loss was reproduced before the fix; manager 4/4 and bridge 2/2 passed after correction. EMU32 graphics delivery passes, but the flow tests carry the mixed-operation backpressure proof.

</details>

<details>
<summary><strong>19. Retire graphics without resurrecting stale images</strong></summary>

**Why:** Closing or replacing a producer must invalidate its images and prevent delayed work from bringing them back.

**Behavior:** Retirement generations, tombstones, acknowledgements, freeing deletes and single-flight mobile writes coordinate lifecycle. Stale session routes are replaced and graphics are reattached after recovery.

```mermaid
flowchart LR
 A[Producer generation] --> B[Placement and acknowledgement]
 B --> C[Retire or replace]
 C --> D[Tombstone and free images]
 D --> E[Reject late old-generation work]
```

**Tested / limits:** LOCAL: focused host/mobile lifecycle flows passed at component revisions. EMU32 confirms synthetic graphics activity. Historical real-browser growth and the live runner cleanup gap are not resolved by these results.

</details>

<details>
<summary><strong>20. Keep input, zoom and image geometry aligned</strong></summary>

**Why:** A tap must map to the source image, and terminal font changes must agree with the cell dimensions the host sees.

**Behavior:** Cell/pixel dimensions travel through native ClientHello and resize paths. Source-image pointer mapping, terminal zoom and gesture/frame pacing use the relevant geometry and pane identity.

```mermaid
flowchart LR
 A[Font and viewport] --> B[Native cell dimensions]
 B --> C[Host graphics geometry]
 D[Phone touch] --> E[Map to source-image coordinates]
 C --> E
 E --> F[Input for the owning pane]
```

**Tested / limits:** LOCAL: geometry and bridge flows pass. EMU32 retains numeric cell evidence, pane-bound checkerboard pixels and positive delivery. Every physical device, arbitrary image size and cross-platform pointer path is not proven.

</details>

### Keep the herd responsive

<details>
<summary><strong>21. Batch work updates without publishing false lifecycle changes</strong></summary>

**Why:** A title update should change a title; it should not make unrelated state behave as though work started or ended.

**Behavior:** Inbound session frames are merged before store updates. Title-only changes avoid unnecessary lifecycle publication, reducing repeated work while keeping actual session changes visible.

```mermaid
flowchart LR
 A[Incoming session frames] --> B[Batch and merge]
 B --> C{Meaningful change}
 C -->|Title only| D[Update title]
 C -->|Lifecycle| E[Publish work transition]
```

**Tested / limits:** LOCAL: title-contract and lifecycle flows pass; the retention fixture clock is pinned to its historical event window. EMU32 runs a changing herd during measured windows; it is not a claimed universal speedup percentage.

</details>

<details>
<summary><strong>22. Scroll without a React update for every frame</strong></summary>

**Why:** Gesture movement should not force global JavaScript state work at the display refresh rate.

**Behavior:** The live strip tracks movement through refs and commits visible selection at gesture end. Tree/event refresh is paced. Diagnostics compare requested gesture travel with actual movement.

```mermaid
flowchart LR
 A[Continuous gesture] --> B[Native movement and refs]
 B --> C[Gesture end]
 C --> D[Commit visible selection]
 A --> E[Measure requested and actual travel]
```

**Tested / limits:** LOCAL: gesture metrics and component flows pass. EMU32 proves movement and measures herd/document/text/graphics windows. Earlier phone and capacity observations are not substituted for exact-head device evidence.

</details>

### Understand provider usage

<details>
<summary><strong>23. Read OMP and OpenCode activity across profiles</strong></summary>

**Why:** Usage must follow the provider profile actually in use, including non-default local directories.

**Behavior:** OMP local SQLite aggregation handles profile/config paths and relative-path resolution. OpenCode activity contributes recency; controlled ccusage output feeds the common usage view. Reads are bounded and avoid turning provider history into unrestricted prompt export.

```mermaid
flowchart LR
 A[OMP profile data] --> C[Bounded local aggregation]
 B[OpenCode activity] --> C
 C --> D[Common provider usage rows]
 D --> E[Usage screen]
```

**Tested / limits:** LOCAL: current Usage flow and path/profile checks pass. EMU32 runs the real implementation with deterministic HOME data and a ccusage stub; that does not validate the upstream parser or every real account database.

</details>

<details>
<summary><strong>24. Order tabs by recent activity in the correct date window</strong></summary>

**Why:** The provider used most recently should be easy to find; a seven-day view should end today in the local calendar.

**Behavior:** Tabs use activity ordering, retain explicit selection, and show the selected provider totals. The date window follows local calendar days, including daylight-saving boundaries.

```mermaid
flowchart LR
 A[Activity timestamps] --> B[Local seven-day window]
 B --> C[Order providers by recency]
 C --> D[Selected tab]
 D --> E[Matching totals]
```

**Tested / limits:** LOCAL: Usage/date behavior is covered by the current flow. EMU32 requires OMP 150 → OpenCode 300 → OMP 150 with selected-state and ordering assertions; these are fixture totals, not live quota values.

</details>

<details>
<summary><strong>25. Distinguish local tokens from Go quota availability</strong></summary>

**Why:** Unknown remaining quota is not zero usage and should not hide known local activity.

**Behavior:** Go quota/config handling returns supported live limits when available and actionable authentication/backend failure otherwise. Valid empty/null JSON overrides suppress disk configuration; invalid JSON falls back according to the configuration contract.

```mermaid
flowchart TD
 A[Local activity] --> D[Known token usage]
 B[Go private configuration] --> C{Quota request outcome}
 C -->|Available| E[Show limit and remaining]
 C -->|Unavailable| F[Explain required action]
 D --> G[Usage view]
 E --> G
 F --> G
```

**Tested / limits:** LOCAL: configuration/account-switch and policy flows pass. EMU32 explicitly requires the actionable Go unavailable message. Successful live Go entitlement/quota is not claimed.

</details>

<details>
<summary><strong>26. Keep cached usage scoped to the account</strong></summary>

**Why:** Switching accounts must not reuse another account’s limits, and a transient failure must not become durable truth.

**Behavior:** Credential-sensitive cache identity uses scrypt rather than raw credentials. Failed fetches are not cached as successful durable results; subprocess/parsing work remains bounded.

```mermaid
flowchart LR
 A[Current account input] --> B[Scrypt cache identity]
 B --> C[Account-scoped lookup]
 C --> D{Fetch result}
 D -->|Success| E[Cache valid data]
 D -->|Failure| F[Return unavailable without success cache]
```

**Tested / limits:** LOCAL: Usage account/configuration flows and independent policy review passed. Fixtures validate isolation behavior without exporting credentials.

</details>

### Configure realtime voice

<details>
<summary><strong>27. Choose one provider and preserve that choice</strong></summary>

**Why:** Changing a provider should happen in one place and survive subsequent setup.

**Behavior:** One Realtime voice picker exposes Codex Voice, Grok, Gemini Live and OpenAI Realtime. Codex is the default only when there is no valid saved choice. Consolidated adapters replace separate provider packages.

```mermaid
flowchart LR
 A[Open Realtime voice] --> B{Valid saved provider}
 B -->|Yes| C[Keep saved choice]
 B -->|No| D[Select Codex default]
 C --> E[Single provider picker]
 D --> E
```

**Tested / limits:** LOCAL: provider flows 11/11 and package default/configuration checks pass. LIVE-PARTIAL: the earlier device run showed all four choices, Codex default, Grok selection and restored Codex.

</details>

<details>
<summary><strong>28. Open the setup screen belonging to the selected provider</strong></summary>

**Why:** A logged-in provider should not be sent to an unrelated API-key form.

**Behavior:** Configure resolves backend-declared configurationContributionId only to an existing navigation screen in the same plugin. Missing/invalid metadata reports unavailable. Codex uses a login-only screen with no key actions.

```mermaid
flowchart TD
 A[Configure selected provider] --> B{Declared screen valid in same plugin}
 B -->|Yes| C[Open declared setup]
 B -->|No| D[Show unavailable]
 C --> E[Codex login or provider-specific setup]
```

**Tested / limits:** LOCAL: voice RPC, manifest parser, bundled guards and packaged RPC/key-security smoke pass. LIVE-PARTIAL: Codex login-only setup and absence of API-key controls were observed.

</details>

<details>
<summary><strong>29. Keep realtime audio native and service-ordered</strong></summary>

**Why:** Continuous conversation requires an active capture service and a streaming audio connection, not merely a configured account.

**Behavior:** The native streaming speech-to-speech path remains in place; microphone foreground service must precede capture. Provider adapters translate their audio/signaling protocols while backend configuration stays out of mobile provider conditionals.

```mermaid
flowchart LR
 A[Start realtime] --> B[Microphone foreground service]
 B --> C[Native capture and streaming]
 C <--> D[Selected provider]
 D --> E[Received speech playback]
```

**Tested / limits:** LOCAL: native WebRTC flow 1/1 and voice RPC lifecycle checks passed. Live microphone, signaling, tool use, returned speech and end cleanup have not yet completed as one accepted run. Login readiness is not audio or entitlement PASS.

</details>

<details>
<summary><strong>30. Choose only a currently live voice target</strong></summary>

**Why:** Remembered activity is a hint, not proof that an agent still exists.

**Behavior:** Fallback and remembered targets are filtered through fresh session.list IDs intersected with Herdr tree pane session IDs. Machine/focus guards remain; malformed nested workspaces, tabs or panes fail closed instead of throwing or selecting stale work.

```mermaid
flowchart LR
 A[Fresh session list] --> C[Intersection of live routes]
 B[Fresh Herdr tree] --> C
 D[Focused or remembered candidates] --> E[Filter by live membership]
 C --> E
 E --> F[Valid target or no target]
```

**Tested / limits:** LOCAL: existing dictation flow 8/8 covers newer stale history, RPC failure and nested malformed entries including null panes. EMU32 includes this correction. The later full spoken interaction remains unverified.

</details>

### Direct work through the assistant

<details>
<summary><strong>31. Inspect useful work metadata before asking for a name</strong></summary>

**Why:** The assistant needs the same work map the user relies on to interpret an incomplete request.

**Behavior:** Context/list tools expose live task, display name, workspace, tab, status, provider kind, promptability and change times, plus desktop focus and current voice/last-tool context. Paged responses keep the initial answer bounded.

```mermaid
flowchart LR
 A[Live work catalog] --> B[Task and status metadata]
 C[Desktop and voice context] --> B
 B --> D[Bounded assistant context]
 D --> E[Useful suggestion or concrete clarification]
```

**Tested / limits:** LOCAL: fa0fa774 host layout/replay/receipt checks and realtime/dictation flows passed; bc90a182 provider/catalog flows pass. New assistant mobile context is newer than EMU32 and needs a matching device run.

</details>

<details>
<summary><strong>32. Find agents with conservative fuzzy names and task descriptions</strong></summary>

**Why:** Spoken references are often approximate, but uncertainty must not send work to the wrong agent.

**Behavior:** Shared matching prefers exact/prefix names, then conservative similarity and token matching over names/tasks. Ambiguous matches yield concrete choices. An omitted recipient sends nothing.

```mermaid
flowchart TD
 A[Spoken name or task] --> B[Search fresh candidates]
 B --> C{Unique confident match}
 C -->|Yes| D[Resolve target]
 C -->|No| E[Offer named choices]
 E --> F[Wait for clarification]
```

**Tested / limits:** LOCAL: shared matching is exercised through existing realtime/coordinator flows. Real speech recognition, fuzzy navigation and ambiguity handling on the final mobile source still need device evidence.

</details>

<details>
<summary><strong>33. Use phone focus, recent views and semantic navigation</strong></summary>

**Why:** What is focused on the desktop and what the user last viewed on the phone are different pieces of context.

**Behavior:** The app tracks bounded recent route visits for the current process and exposes up to five recently viewed live agents after fresh membership checks. Semantic navigation can open a uniquely resolved agent; desktop pane focus remains a separate action.

```mermaid
flowchart LR
 A[Phone route visits] --> B[Bounded recent-view history]
 B --> C[Filter against live sessions]
 D[Desktop focus] --> E[Assistant context]
 C --> E
 E --> F[Open phone route or focus desktop explicitly]
```

**Tested / limits:** LOCAL: navigation/context flows pass. History is not persisted across app restarts. EMU32 predates these mobile changes; recency suggestions and actual navigation need fresh device verification.

</details>

<details>
<summary><strong>34. Send complete prompts and report delivery honestly</strong></summary>

**Why:** Creating an agent and delivering its instructions are separate operations with separate outcomes.

**Behavior:** Agent creation can carry the full initial prompt instead of only a short title. Targeting uses resolved live context; receipts distinguish created, queued/delivered and unconfirmed failure. Failure does not trigger automatic re-creation.

```mermaid
flowchart LR
 A[Instruction and resolved target] --> B[Create or select agent]
 B --> C[Send complete prompt]
 C --> D[Report creation receipt]
 C --> E[Report delivery receipt or uncertainty]
```

**Tested / limits:** LOCAL: real host coordinator/layout receipt checks and provider flows pass. No claim of final-device spoken prompt delivery. Arbitrary shell, destructive termination and workspace deletion were not added by this assistant slice.

</details>

<details>
<summary><strong>35. Read real output to summarize ongoing work</strong></summary>

**Why:** A useful summary must be grounded in what the agent actually produced.

**Behavior:** Working agents provide bounded visible output; settled agents can provide bounded recent history. Reads have line/byte bounds and redaction. The assistant must answer using that evidence or explain its scope, rather than substitute status alone.

```mermaid
flowchart LR
 A[Resolved live agent] --> B{Work state}
 B -->|Working| C[Bounded visible output]
 B -->|Settled| D[Bounded recent history]
 C --> E[Grounded summary with scope]
 D --> E
```

**Tested / limits:** LOCAL: bc90a182 uses the real host coordinator in the expanded provider flow and verifies actual work content. The provider’s final transcript is simulated there; live spoken summary completion is still pending.

</details>

<details>
<summary><strong>36. Share tool execution and completion behavior across providers</strong></summary>

**Why:** Provider choice should change wire/audio translation, not the assistant’s work policy or tool reliability.

**Behavior:** Codex, Grok, Gemini and OpenAI use one voice tool runtime/catalog. It bounds requests and concurrency, deduplicates call IDs, cancels abandoned work, and tracks pending answers. It does not automatically retry mutations. Provider adapters retain protocol translation.

```mermaid
flowchart TD
 A[Codex adapter] --> E[Shared voice tool runtime]
 B[Grok adapter] --> E
 C[Gemini adapter] --> E
 D[OpenAI adapter] --> E
 E --> F[Common work and app tools]
 F --> G[Result and answer lifecycle]
```

**Tested / limits:** LOCAL: provider flows 11/11, build, RPC, architecture, bundled and full package smoke pass at bc90a182. FIXED in d40890f0: the outer watch watchdog is a fixed 291-second ceiling; the host retains the bounded requested watch duration. CodeQL now passes. New adapters must implement this contract; unsupported provider protocols do not work automatically.

</details>

<details>
<summary><strong>37. Turn acknowledgement into a result or explicit failure</strong></summary>

**Why:** “Checking now” is not completion; the assistant must return useful evidence or say why it could not.

**Behavior:** Codex client delegation accepts plain text as a read-only context request instead of demanding JSON. Structured catalog calls still govern actions. Duplicate delivery reuses the result, acknowledgement filler is disabled, pending work stays thinking, and an unanswered result produces timeout feedback.

```mermaid
sequenceDiagram
 participant U as User
 participant P as Provider adapter
 participant K as Shared voice runtime
 participant H as Host coordinator
 U->>P: Summarize this agent
 P->>K: Client delegation
 K->>H: Read live context and output
 H-->>K: Bounded real evidence
 K-->>P: Tool result
 P-->>U: Substantive answer or explicit failure
```

**Tested / limits:** LOCAL: old bridge fails the expanded regression; corrected flow passes through the real coordinator with one read and no mutation despite duplicate delivery. DEPLOYED: bc90a182 backend is installed. Live speech is unverified; the screenshot’s exact delegation payload was not retained, so that individual attempt is not conclusively attributed.

</details>

### Prove behavior locally

<details>
<summary><strong>38. Build an APK with source and native provenance</strong></summary>

**Why:** A passing test is useful only if it identifies the application and patched native dependencies actually installed.

**Behavior:** The local build helper uses a frozen forced install, postinstall/native guards and actual patched-file hashes. APK metadata records source identities, and the gate verifies installed APK bytes by pullback SHA.

```mermaid
flowchart LR
 A[Frozen source] --> B[Private dependency install and native guard]
 B --> C[APK and provenance]
 C --> D[Install on owned emulator]
 D --> E[Pull APK back and compare SHA]
```

**Tested / limits:** EMU32: native guard, APK/source/native/harness freeze and pullback SHA pass. Earlier mixed-native builds are retained as invalid evidence. This proves provenance for that build, not the later assistant mobile source.

</details>

<details>
<summary><strong>39. Exercise actual feature mounting end to end</strong></summary>

**Why:** A tap or screenshot alone does not establish that a plugin mounted or an action reached the real implementation.

**Behavior:** The local gate pairs the native app through the real host/relay/crypto stack with isolated Herdr topology, real Git fixtures and real plugin code. It requires specific content, selected states, bounds, source-line placement, graphics pixels and delivered frames.

```mermaid
flowchart LR
 A[Controlled topology and usage fixtures] --> B[Real host and encrypted transport]
 B --> C[Native APK]
 C --> D[Actions and mounted-content assertions]
 D --> E[Retained XML screenshots and report]
```

**Tested / limits:** EMU32: full gate PASS with failures empty; includes viewer cold/warm/offline/recovery, Usage 150/300/150 and graphics. The ccusage stub, synthetic activity and fake catalog are explicit boundaries; no live provider quotas are invented.

</details>

<details>
<summary><strong>40. Measure performance with bounded, inspectable evidence</strong></summary>

**Why:** A responsive-looking frame is not a performance measurement; a sample needs duration, counters and feature continuity.

**Behavior:** Strict before/after mounts surround timestamped CPU/PSS/frame sampling. Raw counts, missing samples, fatal/update-depth checks and delivered-frame requirements accompany gesture evidence. Exact deferred prompts are handled only before sampling.

```mermaid
flowchart LR
 A[Stabilize startup and prove mount] --> B[Bounded measured window]
 B --> C[CPU PSS frames and gesture travel]
 C --> D[Strict post-window feature check]
 D --> E[Report with raw evidence]
```

**Tested / limits:** EMU32: four 32.8-second windows passed: herd JS 4.9%, document 18.2%, text 7.2%, graphics 7.2%. These are emulator observations, not a general speedup or phone benchmark. Capacity/sweep tooling exists; not every capacity is accepted on device.

</details>

<details>
<summary><strong>41. Separate real voice/browser probes from fixture acceptance</strong></summary>

**Why:** Synthetic rendering and configured accounts cannot prove a real web page or a spoken exchange.

**Behavior:** A separate local runner requires matching full-gate provenance and authenticated connectivity. It isolates identity/state without copying auth, bounds workload phases, captures sanitized ADB failures, and uses voice-start to arm the deadline before the tap. Browser ownership uses scoped key/tab retirement and inventory absence.

```mermaid
flowchart TD
 A[Matching full-gate report] --> B[Authenticated real connection]
 B --> C[Bounded voice attempt]
 B --> D[Owned browser attempt]
 C --> E[Actual audio and end evidence required]
 D --> F[Body pixels and retirement evidence required]
```

**Tested / limits:** LIVE-PARTIAL: connectivity and provider settings were observed, but no completed speech exchange or GitHub body-paint acceptance. FIXED in d40890f0: an unproven browser retirement keeps cleanup false and retains lock/scratch. The extracted real finish flow checks both retained-failure and proven-absence paths. Real browser paint remains unverified.

</details>

<details>
<summary><strong>42. Preserve failures and keep fixtures deterministic</strong></summary>

**Why:** A rerun must not overwrite the evidence that explained the bug, and fixtures should not expire with the calendar.

**Behavior:** Unique watched evidence directories, scoped child cleanup and emulator locks accompany local runs. XML attributes decode once; lifecycle fixtures pin time within retention; inlay fixtures use committed content; graphics tests await the real worker rather than a fixed number of event-loop turns.

```mermaid
flowchart LR
 A[Unique run directory] --> B[Deterministic existing flows]
 B --> C[Retain pass or failure evidence]
 C --> D[Await scoped cleanup]
 D --> E[Release device lock when safe]
```

**Tested / limits:** LOCAL: lifecycle 6/6, inlay flow, gesture/entity checks and graphics ordinary/delayed 2/2 passed at their focused revisions; final repository suite passes. The remaining browser cleanup exception was closed in d40890f0 by retaining lock/scratch unless absence was proven.

</details>

### Ship an understandable package

<details>
<summary><strong>43. Package and deploy the consolidated release</strong></summary>

**Why:** The installed package should contain the same host, plugin and UI resources that were built and checked.

**Behavior:** Version 0.1.26 consolidates component work into this release PR. Fresh packing includes web resources and plugin files; installed-package smoke verifies real RPC/configuration/security behavior. The local backend now uses a stable bc90a182 snapshot.

```mermaid
flowchart LR
 A[Consolidated release source] --> B[Build and pack]
 B --> C[Installed-package smoke]
 C --> D[Local host relay and bundled roots]
 D --> E[Runtime readiness check]
```

**Tested / limits:** LOCAL: fresh pack including web export and full package smoke pass. DEPLOYED: service restart at 04:33:02 UTC, matching host/relay/plugin roots and open relay. Main is not merged and no release is published; local readiness is not live-audio acceptance.

</details>

<details>
<summary><strong>44. Explain voice data use and supply store assets</strong></summary>

**Why:** Permission text and store material must describe what the product actually does.

**Behavior:** Microphone disclosure explains realtime audio sharing with the selected provider, distinct from on-device dictation. Approved iPhone/iPad raw and composed screenshot assets are included with release/setup documentation.

```mermaid
flowchart LR
 A[Actual audio behavior] --> B[Permission and privacy wording]
 C[Product screenshots] --> D[iPhone and iPad release assets]
 B --> E[Release materials]
 D --> E
```

**Tested / limits:** SOURCE: wording/assets are present in the diff. Build/package checks pass; this does not constitute store submission, approval or a new live audio validation.

</details>

<details>
<summary><strong>45. Remove obsolete surfaces and retain scoped maintenance updates</strong></summary>

**Why:** A consolidated release should have one supported path per capability and a truthful dependency inventory.

**Behavior:** Separate voice adapters become one bundle; unused example/inbox/server bundles, Runbook and the dedicated broken Preview UI are removed. Underlying preview-tunnel seams remain. Net maintenance includes Android action pins and xmldom/demo fast-uri updates; five other dependency PRs remain separate.

```mermaid
flowchart LR
 A[Overlapping or obsolete surfaces] --> B[Consolidated supported paths]
 C[Actual final dependency diff] --> D[Scoped maintenance changes]
 B --> E[Release package]
 D --> E
```

**Tested / limits:** LOCAL: final repository suite, architecture, bundle-pruning and package smoke pass. Preact, Expo UI, worklets, uuid and Zustand upgrades are not claimed: their final resolutions retain existing versions despite historical commits.

</details>

### Latest downloadable artifacts

[Build artifacts and provenance](https://github.com/umeranjum17/muxr/releases/tag/untagged-51e4e216a43eebc5c4cf) are attached to a draft build release for owner review. The **ARM64 APK is 0.1.26 / versionCode 355**, signed with the existing local muxr internal-testing certificate. The x86_64 APK is a separate emulator test-signed build. The npm tarball is a downloadable package, not a registry publication. No store rollout or stable release is triggered here.

The fresh d40890f0 full gate passes with no failures: herd JS 4.8%, document 18.2%, text 7.6%, graphics 7.5% over 32.8–32.9 seconds per window. Viewer cold/warm/offline/recovery, Usage selection/recency and graphics evidence passed. Initial local packaging-command and Maestro-path failures remain retained; neither was treated as a product PASS.

The older evidence below remains historical. The new generic assistant mobile source is now included in the passing APK, but the full gate does not independently prove every fuzzy voice-navigation or live spoken action.

### What has actually been tested

| Evidence | Revision and result | What it establishes |
| --- | --- | --- |
| Shared runtime and installed package | **`bc90a182` PASS** | Build, provider flows **11/11**, real coordinator summary regression, voice RPC lifecycle, bundled guards, architecture checks, fresh pack/web export and full package smoke. Local host/relay/bundled roots now match this package. No live spoken-completion claim. |
| Consolidated local checks | **`4b822a91` PASS** | Build; mobile TypeScript; host layout/replay/receipt self-check; provider flow **11/11**; realtime/dictation flows **15/15**; voice RPC lifecycle; architecture and syntax/diff checks. |
| Consolidated CI | **`d40890f0`: suite, both analysis jobs and CodeQL PASS** | Current [PR checks](https://github.com/umeranjum17/muxr/pull/209/checks). The earlier bc90a182 timer alert is closed. CI does not run the local emulator gate. |
| Full local emulator gate | **Earlier `32edfc54` PASS**, `failures: []` | Fresh native-verified APK, install/pullback SHA agreement, frozen source/native/harness identities, and cleanup. Covers mounted herd/text terminal; real Git diff and viewer controls; CJK no-wrap; cold/warm relative line-200 navigation, offline error and reconnection; Usage selection/recency **150 → 300 → 150**; native graphics cell dimensions, checkerboard pixels and delivered frames. |
| Emulator performance windows | **Earlier `32edfc54`**, four **32.8 s** windows | Herd JS **4.9%**, document **18.2%**, text **7.2%**, graphics **7.2%**. CPU/PSS/frame pathology checks passed. These are measurements from that emulator/run, not a general speedup percentage or final-head/phone benchmark. |
| Focused implementation evidence | Pinned component revisions | Ordered graphics manager/bridge flows; generated-key encrypted pairing recovery; fresh/stale/malformed voice-target flows; Usage policy/configuration and packaged-provider fixtures. Final-head suite results should be read alongside these narrower checks. |
| Real provider/device observations | Earlier live attempts; **partial only** | Provider picker/default/switching and Codex login-only setup were observed. Authenticated connectivity was observed. No completed speech/audio exchange or accepted real GitHub browser workload is established. |

The emulator fixture uses real host transport and native rendering, but controlled Herdr topology and synthetic OMP/OpenCode activity. Its ccusage wrapper is an explicit fixture boundary; it neither validates ccusage's upstream parser nor fabricates successful live Go quotas. Missing Go credentials must produce the actionable unavailable state. The newer d40890f0 emulator APK includes the assistant mobile changes; its generic feature gate does not claim live voice-tool acceptance.

### Remaining live acceptance and release-publication work

- [x] Resolve the CodeQL timer alert: all checks pass at d40890f0.

- [ ] Verify live spoken follow-through after the shared runtime fix: the local regression passes, but a matched-device request must now return a substantive summary or explicit failure. Preserve the user screenshot as the original failed evidence.
- [x] Build final-source ARM64 and x86 APKs; the d40890f0 x86 full gate passes. Targeted spoken assistant focus, recency, fuzzy navigation and prompt targeting still need phone evidence.
- [ ] Prove a real realtime session with microphone capture, provider signaling, spoken tool use, received audio and end cleanup. Login readiness, UI controls or tracks alone are insufficient.
- [x] Close the final browser cleanup escape: retain lock/scratch whenever a launched browser lacks absence proof; focused execution of the actual finish flow passes.
- [ ] Prove real GitHub body paint and scoped browser retirement/memory behavior. The retained blank screenshot was an early sample around 4–7 seconds, not the whole 20-second budget. Repeated body evidence within the existing bound is still needed; synthetic checkerboard PASS does not resolve the historical real-browser memory concern.
- [ ] Review final CI and the exact-source artifacts before merge/release. Physical-phone and broader iOS/web acceptance remain outside the passing emulator claim. Viewer scrub labels, full CJK horizontal extent, every compact width, and folded/deletion-collision line targets do not all have device proof.

<details>
<summary>Evidence identity and consolidation scope</summary>

The earlier passing APK at `32edfc54` has SHA256 `ba067d0747542dfa89ea69dd2ce03233eea55c878fc911584616873df093145d`.

- Host source: `e2dda4910402207ffaae509de01ab919b9b7582ad43fd62a799c6e7535b8e2c2`
- Host entry: `2ac54f38efcdab3c84e865ac16990300d8c467e2aac7495000c37e47bc9d7742`
- Harness: `22b9aaaba0767e694fb55f9a70c76911db645027cffcef25583fbec41593d4a1`

Reports, raw logs, APK/provenance, screenshots and archives are retained in both coordinating pane attachment roots under `pr209-32edfc54-*`. Current local checks are in `consolidated-release-checks.log`; it retains an initial wrong-runner invocation and the successful Vitest rerun. Earlier failed device reports remain preserved.

Component PRs #211–#221 are represented in this release branch, with #220/#221 merged during final consolidation. Standalone #210 is closed as superseded: its old-base device run failed in VirtualizedList before Usage, which is not proof of a Usage regression. Earlier plugin/viewer/connection work is also included; obsolete branch commits should not be replayed.

The five still-open automated dependency PRs are separate. Final package resolutions retain the existing preact, Expo UI, worklets, uuid and Zustand versions; their appearance in historical commits does not mean those upgrades ship in this diff. Unrelated worktree/tooling files are excluded. PR #209 is merged into main, build artifacts are attached to a draft release, no stable release/store rollout has been published, and the local backend runs the tested bc90a182 package; deployment is not live voice acceptance.

</details>


umeranjum17:
### APK for your phone

**[muxr-0.1.26-vc57-arm64-v8a.apk](https://github.com/umeranjum17/muxr/releases/download/release-0.1.26-838b94fb-vc57/muxr-0.1.26-vc57-arm64-v8a.apk)** — this branch at `838b94fb`, arm64-v8a, 0.1.26 / versionCode 57, same signing key, upgrades in place.

### Superseded PRs closed

Closed as contained in this branch, each verified with `git merge-base --is-ancestor`: #205, #201, #200, #199, #196, #142, and the five action/tool bumps #207, #206, #204, #203, #202. #195 closed as superseded by #205 — same Android graphics work, an earlier implementation, conflicting file for file with the one that has been load-tested.

Still open on purpose: #180, #179, #181, #105, #4. Their commits rode along but the manifest change is reverted here, because the root `resolutions` block pins those five lower — raising only the app manifest declares a range the install does not satisfy. Adopting them means moving the pins and rebuilding native.

### Review fixes landed since the first pass

| Fix | Commit |
|---|---|
| Lifecycle spec's clock pinned — its fixed August timestamps had aged past the store's 7-day retention, so it was failing on the calendar, on `main` too | `5d5a06c6` |
| Host domain self-check asserted the lifecycle republish this branch removes; now asserts the new contract (enrich the catalog, publish nothing) | `7b908067` |
| Two `console.log` of caught errors removed from release builds; README to 0.1.26; five dependabot ranges reverted to match the pins; `session.stop` skipped with a named reason when herdr links the packaged plugin from another root | `838b94fb` |

That last one was the only real `yarn check` failure: the host only trusts a close-plugin registration rooted in its own plugin directory, and this desk has the packed `dist-npm/` copy linked, so the trust anchor was working and the check was wrong to fail on it.

umeranjum17:
APK rebuilt at the branch head so the artifact matches the code under review: **[muxr-0.1.26-vc58-arm64-v8a.apk](https://github.com/umeranjum17/muxr/releases/download/release-0.1.26-9bd92109-vc58/muxr-0.1.26-vc58-arm64-v8a.apk)** — `9bd92109`, arm64-v8a, versionCode 58, same signing key, upgrades in place. The vc57 build on `release-0.1.26-838b94fb-vc57` is the same product code two non-product commits earlier and stays up as a fallback.

CI is green at `9bd92109`: suite, CodeQL and both analyses pass.

umeranjum17:
APK replaced: **[muxr-0.1.26-vc59-arm64-v8a.apk](https://github.com/umeranjum17/muxr/releases/download/release-0.1.26-9bd92109-vc59/muxr-0.1.26-vc59-arm64-v8a.apk)** (`9bd92109`, versionCode 59).

vc58 is deleted. It was mine, not the branch's: I ran `yarn install --ignore-scripts` to normalise five lockfile keys, which skips `patch-package`, and then built with an ad-hoc script instead of `scripts/buildAndroidLocal.sh`. That left `node_modules` unpatched, so the APK shipped without the Ghostty patch — accessory key bar visible, no scroll-row forwarding, no Kitty snapshot or pointer path — plus the dictation and image-size patches. Confirmed by dex markers: vc57 contains `nativeKittySnapshot` and `onScrollRows`, vc58 contains neither, vc59 contains both.

Nothing in the branch or in CI is affected: `mobile-android-internal.yml` installs with scripts enabled and gates on `node scripts/diagnostics/application/verifyNativePatches.mjs` before signing, and `buildAndroidLocal.sh` runs `patch-package --error-on-fail` plus the same verifier. Only my out-of-band build skipped them.

umeranjum17:
## Graphics panes: measured, then fixed

The text terminal was good; a graphics pane was not. Scrolling a `terminal-browser`
pane from the phone produced nothing for seconds, then jumped a page and a half.
This traces the cause with numbers and fixes it as a general mechanism — nothing
here is specific to that one program.

### What it actually cost, measured against the live desk

A phone control-attach sizes the pane from the phone, so the producer paints
**539×575 = 1.18 MB of RGBA** per frame. Herdr re-encodes program images into the
app-client stream as **uncompressed base64** (`a=t,t=d,f=32` then `a=p`) — about
1.6 MB a frame — and that socket sustains **~3 MB/s**.

```mermaid
flowchart LR
    P["producer<br/>repaints per notch"] -->|"1.6 MB base64<br/>per frame"| H["Herdr"]
    H -->|"~3 MB/s socket"| B["muxr bridge"]
    B -->|"E2EE frame"| R["relay"]
    R --> Ph["phone"]
```

One gesture used to put **9–17 MB** on that socket, because the host turned every
finger row into one wheel notch and sent up to **40 per flush**; each notch is
120 CSS px of smooth scroll in anything Chromium-shaped, and every animation
paint is another full frame. 16–32 MB through 3 MB/s is five to ten seconds of
queue, which is exactly what "it doesn't scroll" felt like. The 40 notches also
explain the overshoot: a ten-row drag scrolled about 1200 px.

### The fix, and why it is producer-agnostic

**A notch is three rows of travel**, which is what a physical wheel delivers for
the same gesture — no producer publishes its own detent, so anything pixel-exact
would be tuned to one program and wrong for the rest.

**The gesture is bounded by frames, not by a count.** At most four notches are in
flight; the rest of a fling is released one notch per delivered frame, with a
100 ms fallback for programs that never repaint and a bounded backlog so a fling
is shortened rather than queued. A pane scrolls exactly as fast as it can answer.

**Newest wins, keyed by surface.** Inline forwarding is now serialized through one
worker, and a placement a later one replaces is dropped before it costs a layout
probe, a base64 decode or a compression. The key is the cell and cell span, not
the image id: a repainter mints a new id every frame, so id-keying would coalesce
nothing, while a legend beside a plot keeps its own slot.

**Three generic bugs fell out of that.** Every frame used to emit `a=d,d=A`, so a
second image in the same pane was erased by the first repaint. Program deletes
carried Herdr's ids while the phone held reallocated ones, so `kitten icat --clear`
never cleared anything. And every image was refitted to fill the pane, so a
four-line inline image was blown up to full screen. Ids are now Herdr's own end to
end, a replacement deletes exactly the image it replaced, and a placement that
does not cover its pane keeps the cell and span Herdr gave it — carried to the
phone as `graphicsSurface`, which is also what gates pointer takeover and the new
optimistic pan.

**The finger moves something immediately.** While a repaint is in flight the phone
translates the delivered pixels with the gesture (native driver, graphics panes
only) and snaps back when the real frame lands, or after the ack window if the
producer never answers.

**Deflate level 6 above 512 kB.** Every byte is decrypted in JavaScript on the
phone, so shrinking a frame 2.5× is worth 1 ms of host CPU.

### Before → after, same page, same desk

| gesture | notches sent | frames | first frame | last frame | bytes |
|---|---|---|---|---|---|
| before, any drag | up to 40 | 10+ queued | — (none in 25 s) | — | 9–17 MB |
| 12-row drag | 4 | 11 | **20 ms** | 192 ms | 0.50 MB |
| 30-row drag | 4 + 6 drained | 14 | **29 ms** | 317 ms | 0.85 MB |
| 60-row fling | 4 + 8 drained | 15 | **24 ms** | 309 ms | 0.87 MB |

Host pipeline, from the first admitted byte to the frame written to the phone:
**p50 6 ms, p95 29 ms**, frame p95 88 kB, one superseded frame dropped per burst.
A 60-row fling moves the page 997 px — about a screen and a half, where it used to
be nearly five.

### It is now part of the load test

`perf/fake-herdr/graphics.mjs` emitted a 32×32 image at 4 Hz, which proved the
path existed and nothing about its cost. It now emits pane-sized changing frames
at the measured geometry, paced at the 3 MB/s the real socket sustains, painting
whichever pane the phone is attached to, and requests frames per parsed SGR wheel
notch the way Chromium repaints. The gate gained a `graphics pane scroll` phase
(90 s of mixed drags and flings) and reads the host's own account:
`graphics.pipeline { frames, superseded, p50Ms, p95Ms, bytesP95, pixelsP95 }`,
identifier-free, with `graphicsPipelineP95Ms` 250 and `graphicsBytesP95` 800 kB as
hard lines. Superseded frames are reported, never gated — the pipeline answers in
6 ms, so a run with nothing stale to drop is the good case.

### Deliberately not done

Placement-only moves (transmit once, then send `a=p` when an image merely moves)
would remove the last redundant pixels for a producer that scrolls an image
without redrawing it. It needs the web decoder to cache images, and for a
repainter — the case that was broken — the pixels change every frame anyway, so
it buys nothing today. Left out on purpose.

Producer-side frame-rate caps are documented rather than coded: `plugins/panes/README.md`
now states what any Kitty producer owes the phone, with `TERMINAL_BROWSER_FPS=10`
as one example rather than a rule about one program.

umeranjum17:
### Gate note: graphics measurement on this emulator

The graphics phase runs and the app stays healthy on it (JS thread 24.6%, no frame stall, 40/40 terminal attaches, nothing rejected), but the host wrote no `graphics.pipeline` account there, and the reason is not a regression: a graphics bridge only opens for a phone that declares cell pixels, and the software-rendered emulator terminal does not. The gate used to fail on the absence, which is a gate that cannot pass by construction.

It now separates the two cases — no metrics is a note that says the cost went unmeasured, metrics with no account is a hard failure — and the account is proven end to end by `node perf/fake-herdr/stack.smoke.mjs`, which drives the real host with a client that does declare them:

```json
{"event":"graphics.pipeline","frames":18,"superseded":0,"p50Ms":38,"p95Ms":43,"bytesP95":2471,"pixelsP95":309925}
```

The numbers that matter for the felt fix came from the real desk against `terminal-browser`, not from the emulator: first frame 20–29 ms, 0.5–0.9 MB per gesture, host pipeline p50 6 ms and p95 29 ms.

umeranjum17:
### The build

**[muxr-0.1.26-vc60-arm64-v8a.apk](https://github.com/umeranjum17/muxr/releases/download/release-0.1.26-5b58f5c9-vc60/muxr-0.1.26-vc60-arm64-v8a.apk)** — [release `release-0.1.26-5b58f5c9-vc60`](https://github.com/umeranjum17/muxr/releases/tag/release-0.1.26-5b58f5c9-vc60), built from `5b58f5c9`, arm64-v8a, versionCode 60, same internal key as vc59 so it upgrades in place. The PR body's APK line now points here; vc59 stays as a fallback.

It carries only the phone half of the graphics work — the optimistic pan, `graphicsSurface` gating for pointer takeover, `d=I` in the web decoder, and the phone-side graphics diagnostics. The host half is live on the machine already and needs no app update.

The viewer legibility rewrite (computed visual rows, virtualization, edge fades, contrast) is not in vc60; it gets its own build when it lands.

umeranjum17:
### This round: viewer surface, gesture accounting, and a gate that can fail honestly

Nine commits since `b759aad9`, all pushed.

| Commit | What |
|---|---|
| `5b58f5c9` | graphics paced by frames; 9–17 MB per gesture → 0.5–0.9 MB |
| `d0479295` | zoom on both pane kinds; optimistic pan deleted |
| `d24580ef` | a folder is a destination, not an `EISDIR` modal |
| `ded78187` | herd strip + tree stop paying JS per scroll frame — measured 1.2% janky, p95 22 ms |
| `8b5da941` | `notchesSent`/`notchesDropped` + cell-size zoom signal |
| `5407736f` | one edge-to-edge reading column; two prototypes deleted |
| `1d44d336` | seven syntax colours; `↳` on wrapped rows removed; 12 dp type |
| `7356bf26` | title-churn smoke tests the contract instead of its opposite |
| `69c6862a` | gesture gate driven by real injection; `openDocument.yaml` reaches Files |
| `b80c68ca` | keep the perf runner out of the vitest sweep |

**Checks** — vitest 47 files / 190 tests pass; tsc clean across host, contract and mobile; `stack.smoke.mjs` 6/6 against a source host in 15 s; `node --test perf` 1/1.

**Build** — [`release-0.1.26-1d44d336-vc81`](https://github.com/umeranjum17/muxr/releases/tag/release-0.1.26-1d44d336-vc81), arm64, versionCode 81, installs in place.

**Not claimed:** document scroll, terminal text fling and graphics pane scroll have no on-device number — those flows never mounted their screens (JS busy 0.9–10.4% against 26–39% for a real gesture), and the selector fix in `69c6862a` is code-verified only. The scope pin was deleted rather than shipped invisible. The appearance pass is half landed, half specified.

umeranjum17:
Review of head `1d44d336afb68d59a0155520c052bf35cb52602c` against main:

Correctness findings:

- **P1 — CI fails on a history-dependent fixture.** `apps/mobile/sources/components/diff/inlay.spec.ts:19–26` runs `git show 5b58f5c9`, but CI uses a shallow checkout. The current suite log confirms three failures with `fatal: bad revision`. Commit representative patch/file fixtures so the tests work in fresh checkouts and after squash merges. Evidence: https://github.com/umeranjum17/muxr/actions/runs/33886978921/job/101069063129
- **P2 — Backpressure drops independent graphics operations.** `apps/host/src/agent/infrastructure/terminalManager.ts:473–497` stores a single `pendingGraphics` frame per attachment. Under socket congestion, an update for another inline placement overwrites the first; a placement can also overwrite a queued delete, leaving stale imagery. Preserve independent placements and ordered deletes through this downstream queue, as the bridge already does.
- **P2 — Source-line links no longer navigate to their target.** `apps/mobile/sources/components/document/CodeSurface.tsx:56` accepts `highlightLine` but never consumes it. The file route still passes the requested line, yet links such as `file?path=…&line=500` open at the top. Restore selection of file mode and navigation to the requested line.
- **P2 — The wrap control is disconnected.** `DocumentViewer` updates `wrapCode`, but drops it before `ReadingBody`; `CodeSurface.tsx:71` always calls `layoutLines(texts, cols)`. Switching wrap off leaves content wrapped while disabling document swiping. Thread the setting into layout and restore horizontal scrolling when wrapping is disabled.

Standards: `apps/mobile/sources/components/code/codeLayout.spec.ts:69–94` adds separate function/edge-case tests despite the repository's flow-test-only guidance. Keep the meaningful layout flow and consolidate only necessary regression assertions.

Validation: read-only code inspection and existing CI logs. No local full-suite or device testing. Explicitly deferred appearance and performance work was not treated as a defect.

umeranjum17:
Consolidated release head 32edfc54: fresh native-verified APK and full local emulator gate PASS (`failures: []`); CI suite and CodeQL PASS. Exact report/archive/APK/provenance/screenshots are exported in both watched roots under pr209-32edfc54-*.

PR #219 is merged into this release branch. Remaining uncommitted harness work was committed and raised as draft #220 with the unresolved browser cleanup blocker documented. Three active helper-agent panes are closed; no further live probes are running. Live Codex audio and real GitHub body/retirement acceptance remain unresolved, so #209 remains draft. Main and release publication are untouched.

PR #213, opened 2026-09-04
fix(terminal): preserve graphics operations under backpressure

umeranjum17:
Socket backpressure previously retained only one pending graphics operation, so an independent inline placement or delete could be silently overwritten. This keeps graphics operations in order, bounds the queue at 64 MiB/128 frames, and closes explicitly on overflow instead of losing protocol operations.

Independent review PASS; existing real terminal manager flow 4/4 and bridge flow 2/2 pass. Includes mixed placements/deletes under congestion. Only two host files change. This component is being consolidated into release PR #209; final combined emulator validation follows there.

### DIFF
--- a/apps/host/src/agent/infrastructure/terminalManager.test.ts
+++ b/apps/host/src/agent/infrastructure/terminalManager.test.ts
@@ -86,6 +86,7 @@ vi.mock('ws', async () => {
 });
 
 import { TerminalManager } from './terminalManager.js';
+import { HerdrGraphicsBridge, type HerdrGraphicsRegistration } from './herdrGraphicsBridge.js';
 
 describe('TerminalManager stream exit', () => {
     beforeEach(() => {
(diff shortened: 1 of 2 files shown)


umeranjum17:
# Graphics backpressure correction — FINAL PASS

Reviewed frozen two-file change in fix-graphics-backpressure atop49cc742d6ac212703e230051ac9fdab2d0170eee. Independently matched SHA256: terminalManager.ts73769c24f18bcbd1ccb80d207bed3100decabd3adb7883260bdc79ed6307d142; terminalManager.test.ts68eb001c7218d77e80ba5b58164ba50ba252cea1212b368d48dfea940b566d79. No concrete blocker; ready for signed commit of reviewed bytes.

Exact baseline methods independently reproduced lost inline placement/delete at600KiB backpressure. Correction queues every bridge output in FIFO order; no unsupported transport coalescing.64MiB/128-frame pending caps checked before enqueue; overflow explicitly closes channel. Drain waits for low watermark, sends until high watermark and retains remainder. Detach/close clears bytes/frames/timer and prevents late flush.

Existing manager flow uses actual attach/graphics registration callback and verifies independent placement/delete order, re-block mid-drain/resume, byte-overflow close and no late flush. Owner4/4 PASS; reviewer inspected entire production/test diff and diff --check PASS. Bridge remains authoritative for per-placement coalescing. Additional bridge flow is supplementary, not needed to waive a known concern. No production auth/guard changes, builds/device/integration edits or new test suite by reviewer.

PR #219, opened 2026-09-05
fix: select realtime voice targets from fresh live sessions

umeranjum17:
The global realtime voice shortcut could select a newer historical session from mobile storage even when that route no longer existed on the host, producing “That agent is no longer available.”

Restrict remembered and fallback targets to the intersection of fresh host session inventory and tree pane membership. Preserve focus and machine checks, and return no target when fresh RPC data is unavailable or malformed, including null pane entries.

Validation: existing dictation flow 8/8 passed; independent review of the corrected two-file change passed. Native voice transport and provider adapters are unchanged. This mobile change still requires a fresh consolidated APK/full emulator gate and live voice validation; no audio or entitlement PASS is claimed.

Consolidation target: #209. Signed commits 75548353 and 02050909.


umeranjum17:
Independent review and root follow-up inspection closed the null-pane crash in 02050909. Existing dictation flow: 8/8 PASS. Consolidated with the original target correction into release head 32edfc54. Fresh APK/full local gate and live voice validation are pending; this code review does not claim audio acceptance. Main remains unmerged and #209 remains draft.

PR #336, opened 2026-09-17
ci: make workflow manual-dispatch only

umeranjum17:
Per the 2026-09-14 standing rule, muxr must not run GitHub Actions automatically on PRs or main merges; workflows stay manually dispatchable after local testing. `ci.yml` was the last workflow with an automatic trigger (`pull_request: branches: [main]`); it is now `workflow_dispatch`-only, so **pull requests will no longer get automatic checks — this is deliberate**. Validation happens locally (`yarn run check:fast` while iterating, `yarn run check` before opening a PR). The `fast` and `suite` jobs are unchanged and still run when dispatched by hand. The header comment was rewritten to state this policy instead of warning against it, and the concurrency group no longer keys off pull-request-only expressions (`group: ci-${{ github.ref }}`, always cancel-in-progress).

**Consequence needing a human decision:** `suite` is currently configured as a **required status check on main**. With no automatic PR runs, no `suite` check context will ever appear on pull requests, so PRs will be unmergeable until branch protection drops that required check. Repository settings were intentionally not touched here.

Verified: `actionlint` clean, YAML parses with `workflow_dispatch` as the only trigger and both jobs intact, and `gh-axi workflow list` still shows `ci` active.


umeranjum17:
Closing: pull-request tests stay on as the merge gate. Releases and site deploys remain manual-only.

